PR #180, opened 2021-06-30
Minor fixes

Siegrift:
## Includes

* Rename active proposal path from `proposal` to `governance`, because otherwise the navigation button won't be highlighted. This also matches the convention for history page.
* Rename proposals to `active proposals` or `past proposals`
* Minor changes to unsupported network modal

### DIFF
--- a/src/components/sign-in/sign-in.tsx
+++ b/src/components/sign-in/sign-in.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { Fragment, useState } from 'react';
 import { ethers } from 'ethers';
 import WalletConnectProvider from '@walletconnect/web3-provider';
 import Web3Modal from 'web3modal';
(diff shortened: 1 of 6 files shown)


src/components/sign-in/sign-in.tsx
```diff
@@ -194,12 +194,22 @@ const SignIn = ({ dark, position }: Props) => {
           <img className={styles.unsupportedNetworkIcon} src={images.unsupportedNetwork} alt="network not supported" />
           <h5>Unsupported chain!</h5>
 
-          <p className={globalStyles.mtXl}>Supported networks are: {supportedNetworks}</p>
+          <p className={globalStyles.mtXl}>
+            Supported networks are:{' '}
```
andreogle, line 198: I would rather word this as `The following networks are supported:` or just `Supported networks:`

andreogle (approved):
🚀

PR #185, opened 2021-06-30
WIP: voting error handling

Siegrift:
Call static doesn't seem to work for me...

I tried setting the epoch length to one minute just for convenience (in both contract and localhost).
I used one account and I staked some tokens and I am the only staker (so I have absolute majority votin power).
This means every proposal I create automatically executes on creation...
I am trying to play and modify stake target, and I can do that every minute If I disable the `.callstatic` check

However, when the check is enabled I can't create a new proposal and I am getting `API3_HIT_PROPOSAL_COOLDOWN` not sure why... @andreogle do you have any idea why this happens?

### DIFF
--- a/src/contracts/helpers.ts
+++ b/src/contracts/helpers.ts
@@ -56,4 +56,5 @@ export const isZeroAddress = (address: string) => address === constants.AddressZ
 // export const EPOCH_LENGTH = 7 * 24 * 60 * 60;
 
 // Temporary value for user tests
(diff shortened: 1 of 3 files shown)


Siegrift:
We are not sure what's the reason this doesn't work and it has been discussed in https://api3workspace.slack.com/archives/C01N6001YCE/p1625046822244200

Closing for the reason above.

PR #187, opened 2021-07-02
Correctly check genesis epoch when displaying tooltip item

andreogle:


src/pages/proposals/proposals.tsx
```diff
@@ -66,7 +66,7 @@ const Proposals = () => {
   ];
 
   // Only display this message during the genesis epoch
-  if (createNewProposal?.genesisEpochOver) {
+  if (!createNewProposal?.genesisEpochOver) {
```
Siegrift, line 69: I remember reviewing this line, but didn't notice this :/ Shame on me :/

Siegrift (approved):

src/pages/proposals/proposals.tsx
andreogle: Shame on for me implementing it 😅

PR #188, opened 2021-07-02
Add back button and scroll to top on proposal details

andreogle:
### Changes

1. Adds a back button to proposal details and shifts the header layout around a bit (see screenshot). This is the latest design from the Figma.
2. Adds a new button type - `text-link` that doesn't have the green colour.
3. Adds a `useScrollToTop` hook that scrolls to the top when the pathname changes
4. Extracts the `.help` class to global styles as it was duplicated in a few places

### Screenshots

back-button

back-button-mobile


src/components/button/button.tsx
```diff
@@ -5,7 +5,7 @@ import styles from './button.module.scss';
 type Props = {
   children: ReactNode;
   className?: string;
-  type?: 'primary' | 'secondary' | 'link' | 'text';
+  type?: 'primary' | 'secondary' | 'link' | 'text' | 'text-link';
```
Siegrift, line 8: Why not just use type `text`? 

If you dislike the underline, let's override it using component style instead of adding another "generic" button type.

Siegrift (changes requested):

src/pages/proposal-commons/proposal-details/proposal-details.tsx
```diff
@@ -102,17 +105,28 @@ const ProposalDetailsContent = (props: ProposalDetailsProps) => {
...
       </div>
+
       <div className={styles.proposalDetailsHeader}>
-        <p className={styles.proposalDetailsTitle}>{proposal.metadata.title}</p>
+        <div>
+          <h4 className={styles.proposalDetailsTitle}>{proposal.metadata.title}</h4>
+          <div className={styles.proposalTag}>
+            <Tag type={proposal.type}>
+              <span className={globalStyles.capitalize}>
+                #{proposal.voteId.toString()} {proposal.type}
```
Siegrift, line 120: :+1:

Siegrift (commented):
:1st_place_medal: LGTM, but I dislike introducing another version of button

src/components/button/button.tsx
andreogle: Yeah I actually had it like this before. I reverted this to rather override the button styling. Needed a new optional prop though - `btnClassName`

src/components/button/button.tsx
```diff
@@ -5,24 +5,29 @@ import styles from './button.module.scss';
...
   className?: string;
+  btnClassName?: string;
   type?: 'primary' | 'secondary' | 'link' | 'text';
   size?: 'normal' | 'large';
   disabled?: boolean;
   onClick?: () => void;
 };
 
-const Button = ({ children, disabled, type = 'primary', size = 'normal', onClick, className }: Props) => {
+const Button = ({ children, disabled, type = 'primary', size = 'normal', onClick, className, btnClassName }: Props) => {
```
Siegrift, line 15: You don't need this actually. You can make it styled using the following CSS (using the className for "wrapper" element)

```css
.someStyles {
    button {
         ...
    }
}
```

Siegrift (approved):
LGTM now! :1st_place_medal: gj

PR #190, opened 2021-07-05
Implement discussion link

Siegrift:
https://api3workspace.slack.com/archives/C020RCCC3EJ/p1625224629006600


Siegrift:
@andreogle I just realized this is not according to designs, but I don't think it's a hude deal.

andreogle (approved):
I prefer the Figma design to this implementation, but I'm also fine with getting this in as is

Issue #84, opened 2021-06-08
Token amount modal onConfirm error handling


Siegrift:
I find it a bit strange that the token amount modal handles errors in `onConfirm` function. When reading the `onCnfirm` function logic I am not sure whether the error is handled or not (unless I check the modal implementation).

Also, currently we don't close the modal after the action is triggered, but rather when it's completed... I think we agreed to have some loading spinner there - but I am not sure if we decided to wait on design or just use a common spinner. @andreogle do you remember?

Issue #41, opened 2021-05-18
Implement validation


Siegrift:
Handle validation errors or prevent them - will probably need to be finalized by designs first
Related conversation in slack: https://api3workspace.slack.com/archives/C020RCCC3EJ/p1621348785008100

Related to https://github.com/api3dao/api3-dao-dashboard/issues/28

Issue #28, opened 2021-05-17
Handle blockchain errors


Siegrift:
So far, I have only focused on happy path and blockchain errors are ignored.
This is actually not unrealistic requirement, but errors can still happen.

At least, we should make sure that errors won't make the app crash.

PR #191, opened 2021-07-05
Handle all errors, simplify error handling code

Siegrift:
Implements discussion point 3) in https://api3workspace.slack.com/archives/C020RCCC3EJ/p1625224629006600

Also: 
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/84
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/41
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/28

**Review commit by commit.** 

One of the commits just changes adds the return statement on a single line to make the code a bit shorter and (subjectively) more readable. This however changes the semantics of the code a bit (return nothing vs return something). In case you disagree with that change, I'll drop that commit.

### DIFF
--- a/src/logic/dashboard/hooks.ts
+++ b/src/logic/dashboard/hooks.ts
@@ -15,18 +15,19 @@ export const useLoadDashboardData = () => {
 
     const goStakingData = await go(convenience.getUserStakingData(userAccount));
     if (!isGoSuccess(goStakingData)) {
-      notifications.error({
+      return notifications.error({
         message: messages.FAILED_TO_LOAD_CHAIN_DATA,
         errorOrMessage: goStakingData[GO_ERROR_INDEX],
       });
-      return;
     }
     const stakingData = goStakingData[GO_RESULT_INDEX];
 
     const goAllowance = await go(api3Token.allowance(userAccount, api3Pool.address));
     if (!isGoSuccess(goAllowance)) {
-      notifications.error({ message: messages.FAILED_TO_LOAD_CHAIN_DATA, errorOrMessage: goAllowance[GO_ERROR_INDEX] });
-      return;
+      return notifications.error({
+        message: messages.FAILED_TO_LOAD_CHAIN_DATA,
+        errorOrMessage: goAllowance[GO_ERROR_INDEX],
+      });
     }
     const allowance = goAllowance[GO_RESULT_INDEX];
 
--- a/src/logic/dashboard/hooks.ts
+++ b/src/logic/dashboard/hooks.ts
@@ -2,7 +2,7 @@ import { useCallback } from 'react';
 import { useChainData } from '../../chain-data';
 import { useApi3Pool, useApi3Token, useConvenience, usePossibleChainDataUpdate } from '../../contracts';
 import { notifications } from '../../components/notifications/notifications';
-import { go, messages } from '../../utils';
+import { go, GO_ERROR_INDEX, GO_RESULT_INDEX, isGoSuccess, messages } from '../../utils';
 
 export const useLoadDashboardData = () => {
   const api3Pool = useApi3Pool();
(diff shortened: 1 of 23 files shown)


src/logic/dashboard/hooks.ts
```diff
@@ -13,17 +13,22 @@ export const useLoadDashboardData = () => {
   const loadDashboardData = useCallback(async () => {
     if (!provider || !api3Pool || !api3Token || !convenience || !userAccount) return null;
 
-    const [stakingDataErr, stakingData] = await go(convenience.getUserStakingData(userAccount));
-    if (stakingDataErr || !stakingData) {
-      notifications.error({ message: messages.FAILED_TO_LOAD_CHAIN_DATA, errorOrMessage: stakingDataErr! });
+    const goStakingData = await go(convenience.getUserStakingData(userAccount));
```
Siegrift, line 16: Refactored to `isGoSuccess` pattern for consistency and to make TS more happy (such that I don't need to use `!`)

src/logic/proposals/hooks/get-proposals.ts
```diff
@@ -28,9 +28,7 @@ export const getProposals = async (
     };
   });
 
-  // TODO: load this just once for all proposals and save to state
-  const votingTime = await api3Voting.voteTime();
-
+  const votingTime = EPOCH_LENGTH;
```
Siegrift, line 31: Voting time is not changeable and is equal to `EPOCH_LENGTH` so it's fine to use it as constant.

src/pages/proposal-commons/proposal-list/proposal-status/proposal-status.tsx
```diff
@@ -47,11 +49,21 @@ const ProposalStatus = (props: Props) => {
...
+                message: messages.TX_GENERIC_ERROR,
+                errorOrMessage: goTransaction[GO_ERROR_INDEX],
+              });
+              return;
+            }
             setChainData('Save execute transaction', (state) => ({
-              transactions: [...state.transactions, { type: 'execute', tx }],
+              transactions: [...state.transactions, { type: 'execute', tx: goTransaction[GO_RESULT_INDEX] }],
             }));
-            await voting[proposal.type].executeVote(proposal.voteId);
```
Siegrift, line 54: I guess up until now we double voted???

src/pages/dashboard/forms/confirm-unstake-form.tsx
```diff
@@ -21,16 +20,7 @@ const ConfirmUnstakeForm = (props: Props) => {
   const handleAction = async () => {
     setError('');
 
-    const [err] = await go(onConfirm(amount));
-    if (err) {
-      if (isUserRejection(err)) {
-        notifications.info({ message: messages.TX_GENERIC_REJECTED });
-        return;
-      }
-      setError(messages.TX_GENERIC_ERROR);
```
Siegrift, line 30: There is a small change in behaviour here. Previously, we showed an error in the form as helper text and haven't shown an error notification.

Now, there will be no helper text in modal when the `onConfirm` (a.k.a the transaction) fails. We will show the error notification instead. This makes it consistent with new proposal form as well and the rule is:

Show any validation errors in the modal, but onConfirm error handling is not part of the validation anymore and will trigger a toast message error.

src/logic/proposals/hooks/history-proposals.ts
```diff
@@ -62,17 +62,18 @@ const useLoadHistoryProposals = () => {
...
     }
     const proposals = goResponse[GO_RESULT_INDEX];
 
     setChainData(
       'Load history proposals',
       updateImmutablyCurried((state) => {
         if (!state.proposals) {
-          state.proposals = proposals;
-          return;
+          return (state.proposals = proposals);
```
andreogle, line 76: I'm fine with returning the notification ID but having assignment and returning on the same line is a bit weird imo. I don't know if I've ever seen something like this before

src/pages/dashboard/dashboard.tsx
```diff
@@ -164,30 +167,29 @@ const Dashboard = () => {
...
           maxValue={data?.userUnstaked}
         />
       </Modal>
       <Modal open={openModal === 'unstake'} onClose={closeModal}>
         <TokenAmountForm
           title="How many tokens would you like to unstake?"
           action="Initiate Unstaking"
-          onConfirm={async (value) => {
+          onConfirm={(value) => {
+            // NOTE: We are explicitely not closing modal, because it triggers the confirmation modal
```
andreogle, line 187:
```suggestion
            // NOTE: We are explicitly not closing modal, because it triggers the confirmation modal
```

src/pages/proposal-commons/proposal-list/proposal-status/proposal-status.tsx
andreogle: It does look like the contract checks `vote._executed` so I think this second `executeVote` would fail every time?

src/utils/error-handling.ts
```diff
@@ -0,0 +1,18 @@
+import { notifications } from '../components/notifications/notifications';
+import { go, GO_ERROR_INDEX, GO_RESULT_INDEX, isGoSuccess, isUserRejection } from './generic';
+import { messages } from './messages';
+
+export const handleTransactionError = async <T>(transaction: Promise<T>) => {
```
andreogle, line 5: I like wrapping all transactions with a single helper like this but I find the name a misleading. How about something like `handleTransactionSafe` or something like that?

andreogle (approved):

src/pages/proposal-commons/proposal-list/proposal-status/proposal-status.tsx
Siegrift: Yeah, probably

src/logic/proposals/hooks/history-proposals.ts
Siegrift: :) Remind me of my competitive programming career. It's actually an antipattern in practice, but I don't mind this much.

But since you consider it wierd, I'll revert to `return` on single line.

src/utils/error-handling.ts
Siegrift: I think the name is good enough. It handles transaction error (thus the reasoning for this name).

Issue #203, opened 2021-07-08
Allow ENS names in delegation address field


Siegrift:
Based on https://api3workspace.slack.com/archives/C020RCCC3EJ/p1625678335010000

We can use ethers to resolve a name which will make it easier for people to delegate (if the delegate uses ENS name)

bbenligiray:
We should also attempt to [look up](https://docs.ethers.io/v5/api/providers/provider/#Provider-lookupAddress):
- Current delegate for the user
- Proposal creator

PR #205, opened 2021-07-08
Resolve ENS names and addresses, fix e2e tests

Siegrift:
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/203

I think I am looking up the ENS records for all places necessary, but have thought about what I might have missed.

### DIFF
--- a/src/logic/proposals/hooks/get-proposals.ts
+++ b/src/logic/proposals/hooks/get-proposals.ts
@@ -1,16 +1,16 @@
-import { BigNumber } from 'ethers';
+import { BigNumber, providers } from 'ethers';
 import { ProposalType, VoterState } from '../../../chain-data';
-import { Api3Voting, Convenience } from '../../../generated-contracts';
+import { Convenience } from '../../../generated-contracts';
 import { Proposal } from '../../../chain-data';
 import { decodeMetadata } from '../encoding';
-import { blockTimestampToDate } from '../../../utils';
+import { blockTimestampToDate, go, GO_RESULT_INDEX, isGoSuccess } from '../../../utils';
 import { EPOCH_LENGTH, HUNDRED_PERCENT, isZeroAddress } from '../../../contracts';
 import { StartVoteProposal, VOTING_APP_IDS } from './commons';
 
 const toPercent = (value: BigNumber) => value.mul(100).div(HUNDRED_PERCENT).toNumber();
 
 export const getProposals = async (
-  api3Voting: Api3Voting,
+  provider: providers.Web3Provider,
   convenience: Convenience,
   userAccount: string,
   startVoteProposals: StartVoteProposal[],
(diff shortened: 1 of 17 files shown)


src/pages/proposals/forms/delegate/delegate-form.tsx
```diff
@@ -15,29 +15,35 @@ import { handleTransactionError } from '../../../../utils';
...
-    if (!api3Pool) return;
+    if (!api3Pool || !provider) return;
 
-    if (!utils.isAddress(delegationAddress) || delegationAddress === constants.AddressZero) {
+    const delegationTargetLookupName = await go(provider.resolveName(delegationAddress));
+    const delegationTarget = isGoSuccess(delegationTargetLookupName)
+      ? delegationTargetLookupName[GO_RESULT_INDEX]
+      : delegationAddress;
+
+    console.log(delegationTargetLookupName);
```
bbenligiray, line 36: This was probably temporary

Siegrift:
three
two
one

Btw. this is how it looks like

src/logic/treasury-and-delegation/use-treasury-and-delegation.ts
```diff
@@ -32,12 +32,16 @@ export const useTreasuryAndDelegation = () => {
         });
       }
 
+      const delegate = isZeroAddress(data.delegate) ? null : data.delegate;
+      const goDelegateName = delegate ? await go(provider.lookupAddress(delegate)) : null;
```
andreogle, line 36: Kinda weird seeing the await the promise or just returning null as a ternary 😄

bbenligiray:
This should read "Enter address or ENS name here" 
https://github.com/api3dao/api3-dao-dashboard/blob/ens-names-lookup/src/pages/proposals/forms/delegate/delegate-form.tsx#L73

src/pages/proposals/forms/delegate/delegate-form.tsx
```diff
@@ -15,29 +15,33 @@ import { handleTransactionError } from '../../../../utils';
...
   const [error, setError] = useState('');
   const [delegationAddress, setDelegationAddress] = useState('');
   const api3Pool = useApi3Pool();
 
   const onDelegate = async () => {
-    if (!api3Pool) return;
+    if (!api3Pool || !provider) return;
 
-    if (!utils.isAddress(delegationAddress) || delegationAddress === constants.AddressZero) {
+    const delegationTargetLookupName = await go(provider.resolveName(delegationAddress));
```
andreogle, line 31: Can we skip this if [utils.isAddress(delegationAddress)](https://docs.ethers.io/v5/api/utils/address/#utils-isAddress)? i.e. if they entered a valid Ethereum address, there is no need to do a reverse lookup like this

andreogle: Actually, do we even need the lookup here? From the docs:

```
In the ethers API, nearly anywhere that accepts an address, an ENS name may be used instead, which can simplify code and make reading and debugging much simpler.
```

src/logic/proposals/hooks/get-proposals.ts
```diff
@@ -1,16 +1,16 @@
...
 import { decodeMetadata } from '../encoding';
-import { blockTimestampToDate } from '../../../utils';
+import { blockTimestampToDate, go, GO_RESULT_INDEX, isGoSuccess } from '../../../utils';
 import { EPOCH_LENGTH, HUNDRED_PERCENT, isZeroAddress } from '../../../contracts';
 import { StartVoteProposal, VOTING_APP_IDS } from './commons';
 
 const toPercent = (value: BigNumber) => value.mul(100).div(HUNDRED_PERCENT).toNumber();
 
 export const getProposals = async (
-  api3Voting: Api3Voting,
```
andreogle, line 13: Nice catch. I didn't see this was unused

src/pages/proposals/forms/delegate/delegate-form.tsx
Siegrift: Not sure, but I'd rather be explicit about this. I however forgot to rename the variable from `delegationTargetLookupName` to `delegationTargetResolvedName`. I think it's more readable now.

src/logic/treasury-and-delegation/use-treasury-and-delegation.ts
Siegrift: It is what it is :D

andreogle (approved):

src/pages/proposals/forms/delegate/delegate-form.tsx
bbenligiray: I assumed that wasn't the case because [`isAddress`](https://docs.ethers.io/v5/api/utils/address/#utils-isAddress) doesn't recognize ENS names but this looks right. Is there any harm in being explicit about it though?

PR #211, opened 2021-08-02
Add "no console" and "prefer const" rule

Siegrift:
This has sometimes come up in reviews and I've wrote it in my TODO list for later.

Later is now I guess.

### DIFF
--- a/dev-scripts/.eslintrc.json
+++ b/dev-scripts/.eslintrc.json
@@ -1,0 +1,7 @@
+{
+  "extends": ["../.eslintrc.json"],
+  "rules": {
+    "prefer-const": "error",
+    "no-console": ["error", { "allow": ["info", "error"] }]
+  }
+}
--- a/src/chain-data/context.tsx
+++ b/src/chain-data/context.tsx
@@ -14,6 +14,7 @@ const ChainDataContextProvider: React.FC = ({ children }) => {
 
         if (process.env.NODE_ENV === 'development') {
           console.group(reason);
+          // eslint-disable-next-line no-console
           console.info(updatedChainData);
           console.groupEnd();
         }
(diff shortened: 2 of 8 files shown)


dev-scripts/.eslintrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": ["../.eslintrc.json"],
+  "rules": {
```
andreogle, line 3: I don't think these rules are needed if they're defined in the base. This should be enough (?)

```json
{
  "extends": ["../.eslintrc.json"]
}
```

src/chain-data/context.tsx
```diff
@@ -14,6 +14,7 @@ const ChainDataContextProvider: React.FC = ({ children }) => {
 
         if (process.env.NODE_ENV === 'development') {
           console.group(reason);
+          // eslint-disable-next-line no-console
```
andreogle, line 17: Can't these `eslint-disable-next-line` statements be removed since `console.info` been added as an "allow" exception?

andreogle (changes requested):

dev-scripts/.eslintrc.json
andreogle: Nevermind, I missed the difference in `allow`. Maybe just allow them all in the base?

Siegrift: Nah, I want to be strict about what can be used in client (ideally no console).

I can even forbid the console group, and add inline eslint comment if you prefer that.

andreogle: I'm fine with it then 👍🏻

andreogle (approved):

PR #212, opened 2021-08-02
Fix vote radio button

Siegrift:
There was one more bug similar to https://github.com/api3dao/api3-dao-dashboard/commit/ec3207f73851e7cb828308b72e7e516942ab9104 

There are no more (I checked).

### DIFF
--- a/src/pages/proposal-commons/proposal-details/vote-form/vote-form.tsx
+++ b/src/pages/proposal-commons/proposal-details/vote-form/vote-form.tsx
@@ -25,14 +25,14 @@ const VoteForm = (props: Props) => {
             label={<h4>For</h4>}
             onChange={() => setChecked('for')}
             checked={isVotingFor}
-            name="vote"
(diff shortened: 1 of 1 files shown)


src/pages/proposal-commons/proposal-details/vote-form/vote-form.tsx
```diff
@@ -25,14 +25,14 @@ const VoteForm = (props: Props) => {
             label={<h4>For</h4>}
             onChange={() => setChecked('for')}
             checked={isVotingFor}
-            name="vote"
+            name="for"
```
Siegrift, line 28: This is pretty annoying and for no good reason... I think I will remove this param from props and generate unique name inside the RadioButton component later.

andreogle (approved):
Thanks for fixing this so quickly 👍🏻 . I should have also checked myself for more `<RadioButton />` components

Issue #224, opened 2021-08-11
While reading proposal parameters, ENS names of the address types should be reverse-looked up


bbenligiray:
`0xd9dAA0330920006dea50D0aDA4EBc9752d434111` should read `api3-core-tech.eth` instead

![Screenshot 2021-08-10 at 19-42-29 API3 DAO Proposal 6 (1)](https://user-images.githubusercontent.com/19530665/129029180-2e4ae36c-cee3-43d4-883b-2e79e758a210.png)

Issue #223, opened 2021-08-11
While entering proposal parameters, ENS names should be usable for address types


bbenligiray:
This should work

![Screenshot 2021-08-10 at 17-43-43 API3 DAO Governance](https://user-images.githubusercontent.com/19530665/129028993-70c77629-cf61-44e5-9e9e-296744d65773.png)

PR #227, opened 2021-08-14
Ens parameters

Siegrift:
Fixes https://api3workspace.slack.com/archives/C02ALA11APM/p1628590492006400 

which was requested https://forum.api3.org/t/ens-name-for-proposal-creator/289

Fixes https://github.com/api3dao/api3-dao-dashboard/issues/224
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/223

### DIFF
--- a/src/logic/proposals/encoding.ts
+++ b/src/logic/proposals/encoding.ts
@@ -1,5 +1,5 @@
-import { BigNumber, utils } from 'ethers';
-import { ProposalMetadata, ProposalType } from '../../chain-data';
+import { BigNumber, providers, utils } from 'ethers';
+import { DecodedEvmScript, ProposalMetadata, ProposalType } from '../../chain-data';
 import { Api3Agent } from '../../contracts';
 import { errorFn, GoResult, goSync, GO_RESULT_INDEX, isGoSuccess, successFn } from '../../utils';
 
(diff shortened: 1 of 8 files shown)


Siegrift:
I am so happy that I wrote the e2e test because I introduced a hard to spot bug while implementing this.
Check out last commit to see it.

src/logic/proposals/encoding.ts
```diff
@@ -170,14 +180,21 @@ export const decodeEvmScript = (script: string, metadata: ProposalMetadata): Dec
...
+        return ensName || param;
+      })
+    );
 
     return {
       targetAddress: targetContractAddress,
       value,
-      rawParameters,
-      parameters: stringifyBigNumbersRecursively(rawParameters),
+      parameters: stringifyBigNumbersRecursively([...parameters]),
```
andreogle, line 197: The "destructuring to enforce Array shape" comment got lost. Is it still necessary to spread here vs just having `stringifyBigNumbersRecursively(parameters)`?

src/pages/proposals/forms/new-proposal-form.tsx
```diff
@@ -194,7 +196,8 @@ const NewProposalForm = (props: Props) => {
               title,
             };
 
-            if (!validateForm(formData)) {
+            const containsError = await validateForm(formData);
+            if (!containsError) {
```
andreogle, line 200: Nice catch with the subtle bug

andreogle (dismissed):

src/logic/proposals/encoding.ts
Siegrift: I don't think we need that.

andreogle (approved):

PR #231, opened 2021-08-15
Minor styling issues, change proposal details page texts

Siegrift:


src/components/layout/layout.module.scss
```diff
@@ -6,31 +6,42 @@
...
+  display: flex;
+
+  // Make sure the texture is behind the content
   position: relative;
   z-index: 1;
-  max-width: 1020px;
+}
+
+.main {
+  max-width: min(calc(100vw - 170px), 1024px);
```
Siegrift, line 18: The `max-width` attribute is calculated imperatively to avoid the content overflowing the `flex` container, which results in a horizontal scrollbar and "white column" with overflown content.

(Picture taken from https://api3.eth.link/#/history/secondary-5, should not happen anymore)
prod

src/components/layout/layout.tsx
```diff
@@ -8,6 +8,7 @@ import ExternalLink from '../external-link';
 import Button from '../button';
 import ErrorReportingNotice from './error-reporting-notice';
 import classNames from 'classnames';
+import { DesktopMenu } from '../menu';
```
Siegrift, line 11: I needed to split the menu because the texture was overlapping it and z-index trick didn't work anymrore.

src/pages/proposal-commons/proposal-details/proposal-details.tsx
```diff
@@ -72,11 +70,12 @@ const ProposalDetailsContent = (props: ProposalDetailsProps) => {
   const { transactions, setChainData } = useChainData();
   const voting = useApi3Voting();
 
-  const evmScriptData = decodeEvmScript(proposal.script, proposal.metadata);
+  // TODO: implement proper "not signed in" and "does not exist" pages
+  if (!voting || proposal === 'user not signed in')
```
Siegrift, line 74: Blocked https://api3workspace.slack.com/archives/C02ALA11APM/p1629031354024900?thread_ts=1628726176.010900&cid=C02ALA11APM

src/pages/proposal-commons/proposal-details/proposal-details.tsx
```diff
@@ -173,9 +172,9 @@ const ProposalDetailsContent = (props: ProposalDetailsProps) => {
         }
         content={
           <div className={styles.proposalDetailsSummary}>
-            <p className={classNames(styles.proposalDetailsItem, globalStyles.secondaryColor)}>
+            <pre className={classNames(styles.proposalDetailsItem, globalStyles.secondaryColor, styles.scrollX)}>
```
Siegrift, line 175: Before:
before

After:
after

Siegrift: If the content is too long, instead of wrapping let's add a scrollbar... This makes more sense, because people will be pasting URLs or IPFS links and for those it's better when they are not wrapped imo.
scroll

Siegrift: If the content is too long, instead of wrapping let's add a scrollbar... This makes more sense, because people will be pasting URLs or IPFS links and for those it's better when they are not wrapped imo.
scroll

karolko (approved):

andreogle (approved):

Issue #235, opened 2021-08-16
Investigate (and fix) wallet connect issues


Siegrift:
Many wallet connect users complain that they are unable to connect to the DAO dashboard.

PR #236, opened 2021-08-16
Upgrade dependencies and fix wallet connect

Siegrift:
@bbenligiray Since you are using wallet connect can you test whether this works on staging?

The proposed fix for wallet connect is to upgrade the dependency: https://twitter.com/WalletConnect/status/1426226788515078144

UPDATE: The bug was that we were calling `eth_requestAccounts` which is necessary for Metamask, but throws in case of wallet connect.

Fixes https://github.com/api3dao/api3-dao-dashboard/issues/235


andreogle:
The [release notes for v1.6.0](https://github.com/WalletConnect/walletconnect-monorepo/releases/tag/1.6.0) simply say "new default servers" 😅 . Let's hope that helps

src/components/sign-in/sign-in.tsx
```diff
@@ -147,8 +147,10 @@ export const connectWallet = (setChainData: SettableChainData['setChainData']) =
   if (connectionError || !web3ModalProvider) return;
 
   // Wrapped in callback to prevent synchronous error, because `request` property is not guaranteed to exist
-  // TODO: Should we call this in case of wallet connect?
-  const [requestAccountsError] = await go(() => web3ModalProvider.request({ method: 'eth_requestAccounts' }));
+  const [requestAccountsError] = await go(() => {
+    if (web3ModalProvider.isMetaMask) return web3ModalProvider.request({ method: 'eth_requestAccounts' });
```
Siegrift, line 151: @andreogle @bbenligiray I've noticed that this was causing an issue for wallet connect.

I am kinda sad, that I look at the issue sooner, because it was probably our (my) mistake from the start and not just wallet connect flakiness. Though, I am not sure why this suddenly stopped working as it previously did work (when we tested with Santiago)...

Anyway, It's still not 100% reliable though (even with 1.6 update)

andreogle (approved):

Issue #221, opened 2021-08-10
Document that we only support proposals to make a single call on purpose


bbenligiray:
EVMScript supports multiple function calls. For easier UX, we only support proposals to specify a single call. Document this, and how to extend the functionality to have the agent make multiple calls with a single proposal.

Issue #220, opened 2021-08-10
Document EVMScript encoding and verification


bbenligiray:
The dashboard translates proposal parameters into an EVMScript while creating a new proposal. Similarly, the function signature in the metadata is verified using the EVMScript of the proposal. This needs to be documented.

Issue #219, opened 2021-08-10
Document metadata encoding format


bbenligiray:
The dashboard uses encodes new proposal parameters into a `metadata` value of the following form
```
"1|transfer(address,uint256)|API3 DAO BD-API Team Proposal|https://ipfs.fleek.co/ipfs/bafybeicfguu3bfhk3fyz5zn5wlujpksqxsaokse37674fhylouvsqnwd2m" 
```
Similarly, it parses this metadata to populate the Governance & History pages. This encoding/decoding scheme should be documented.

PR #240, opened 2021-08-24
Proposal encoding documentation, split encoding into multiple files

Siegrift:
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/221
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/220
Fixes https://github.com/api3dao/api3-dao-dashboard/issues/219


bbenligiray:
Nice. I was thinking of pure documentation but this is probably more useful.

src/logic/proposals/encoding/encoding.ts
```diff
@@ -1,42 +1,46 @@
+/**
+ * When a proposal is passed, it can be used to execute the EVM script that was specified when the proposal was created.
+ * The script can be a function call or multiple function calls. Proposal needs to be executed by separate transaction
+ * and the executor pays gas also for the EVM script that is run.
```
andreogle, line 4:
```suggestion
 * and the executor will pay the gas cost for the EVM script that is run.
```

src/logic/proposals/encoding/encoding.ts
```diff
@@ -1,42 +1,46 @@
...
+ *
+ * For now, we only support a single function call in the EVM script. The reasons are that it's enough for the most
+ * cases, it's easier to encode it and we can create simpler UI for the users.
```
andreogle, line 7:
```suggestion
 * cases, it's easier to encode and we can create a simpler UI for end users.
```

src/logic/proposals/encoding/encoding.ts
```diff
@@ -46,11 +50,21 @@ type EncodedEvmScriptError = {
...
+ * @param provider a provider which is able to resolve ENS names
+ * @param formData the proposal data to be encoded
+ * @param api3Agent the addresses of the voting app agents
+ */
 export const encodeEvmScript = async (
   provider: providers.Provider,
   formData: NewProposalFormData,
   api3Agent: Api3Agent
 ): Promise<EncodedEvmScript> => {
+  // Ensure that the form parameters form a valid JSON
```
andreogle, line 67:
```suggestion
  // Ensure that the form parameters form a valid JSON array
```

andreogle (approved):
Thanks for this! Just some minor nit comments about the way some comments are worded.

PR #244, opened 2021-08-24
Create React error boundary

Siegrift:


Siegrift:
@andreogle The second commit is just for testing to quickly simulate React error and will be removed.

src/app.tsx
```diff
@@ -1,48 +1,96 @@
...
-import NotFoundPage from './pages/not-found';
 
-function App() {
+const ErrorBoundary: FallbackRender = (props) => {
+  const { error } = props;
+
+  // When the ErrorBoundary component is rendered, there must have been uncaught error somewhere in React component. We
+  // want to send this exception to Sentry.
+  useOnMountEffect(() => {
+    Sentry.captureException(error);
```
andreogle, line 25: I assume this respects if the user has opted out of Sentry?

andreogle: Unrelated to this PR, but it looks like `sendToSentry` is not being used anywhere in the app here (unless I'm missing something)

https://github.com/api3dao/api3-dao-dashboard/blob/601490afb9813f3bfd96057d1d3172f2980316a5/src/components/notifications/notifications.tsx#L91-L94

andreogle (approved):

andreogle:
I tested this locally and it works fine. Thanks for the adding the broken button 👍🏻

PR #248, opened 2021-09-20
running dashboard for mainnet in docker

EnormousCloud:
IPFS-hosted connectivity raises many complaints on dashboard functionality

The decentralized way of running dashboard from a local machine can be spread to address these issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+# This Dockerfile builds a mainnet version for API3 DAO dashboard
+
+FROM node:lts-alpine as builder
+RUN apk add --no-cache git
+WORKDIR /usr/src/app
+ADD . .
+COPY src/contract-deployments/mainnet-dao.json ./src/contract-deployments/localhost-dao.example.json
+RUN yarn 
+RUN yarn tsc && yarn build
+
+FROM nginx:alpine
+EXPOSE 80
+COPY --from=builder /usr/src/app/build /usr/share/nginx/html
+CMD ["nginx", "-g", "daemon off;"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,8 @@ FROM node:lts-alpine as builder
 RUN apk add --no-cache git
 WORKDIR /usr/src/app
 ADD . .
-COPY src/contract-deployments/mainnet-dao.json ./src/contract-deployments/localhost-dao.example.json
 RUN yarn 
-RUN yarn tsc && yarn build
+RUN yarn build
 
 FROM nginx:alpine
 EXPOSE 80
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ docker build -t api3-dao-dashboard .
 docker run -d -p7770:80 --name api3-dao-dashboard api3-dao-dashboard
 ```
 
-This will create a API3 dashboard running on http://localhost:7770/ where it is safe to connect your wallet
+This will create a API3 dashboard running on port 7770 of your localhost where it is safe to connect your wallet.
 
 ## Instructions for testing on Rinkeby
 
--- a/README.md
+++ b/README.md
@@ -4,6 +4,18 @@ The implementation of the DAO dashboard.
 
 [![ContinuousBuild](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml/badge.svg?branch=main)](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml)
 
+## Running dashboard on Mainnet in docker container
+
+The decentralized approach of being a DAO member is to run API3 dashboard on your local machine. For that you will just need `git` and `docker`:
+```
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,14 @@
+# This Dockerfile builds a mainnet version for API3 DAO dashboard
+
+FROM node:lts-alpine as builder
+RUN apk add --no-cache git
+WORKDIR /usr/src/app
+ADD . .
+COPY src/contract-deployments/mainnet-dao.json ./src/contract-deployments/localhost-dao.example.json
```
Siegrift, line 7: You actually don't need this line - it will be done as part of `yarn postinstall`

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN yarn 
+RUN yarn tsc && yarn build
```
Siegrift, line 9: You don't need `yarn tsc` as that only runs TS checks

README.md
```diff
@@ -4,6 +4,18 @@ The implementation of the DAO dashboard.
 
 [![ContinuousBuild](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml/badge.svg?branch=main)](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml)
 
+## Running dashboard on Mainnet in docker container
+
+The decentralized approach of being a DAO member is to run API3 dashboard on your local machine. For that you will just need `git` and `docker`:
+```
+git clone --depth=1 git@github.com:api3dao/api3-dao-dashboard.git
+cd api3-dao-dashboard
+docker build -t api3-dao-dashboard .
```
Siegrift, line 13: Maybe we could push the image on api3 docker hub (https://hub.docker.com/u/api3) to avoid the need to clone the repo beforehand?

Siegrift (commented):
Hi :), thanks for doing this - I think we should make running DAO dashboard locally simpler

> IPFS-hosted connectivity raises many complaints on dashboard functionality
The decentralized way of running dashboard from a local machine can be spread to address these issues.

Yeah, we mention those issues in `Updating the name servers` (which is not the best place for this), but we never mention that running locally is a viable option too.

That said, this approach seems a bit complex to me, since to run the dao-dashboard locally you just need to do `yarn && yarn start` after cloning the repo - maybe it's enough to mention this at the top of the README?

Also, I am not sure how many "regular" users do indeed want to run this locally - as downloading git and docker can be painful.

@bbenligiray @andreogle thoughts?

README.md
andreogle: Yeah that might be nice

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+FROM nginx:alpine
```
andreogle, line 11: Do we need nginx for this? Can't we just expose port 3000 from the create-react-app server? It's wouldn't be using the output html/js files, but I don't think that's a problem for this?

andreogle (commented):
I agree with @Siegrift that running the DAO dashboard locally is as simple as running `yarn && yarn start`. This allows for connecting to Ethereum mainnet out of the box, so I wonder how much of an issue this is (?)

That said, I have no problem with adding a Dockerfile and it might be nice to push it to Dockerhub so it's available to anyone with Docker installed.

README.md
EnormousCloud: yes, this is a valuable idea

Dockerfile
EnormousCloud: nope, post install will copy your local setup. There is a trick of how to connect to mainnet. You need to study source

Dockerfile
EnormousCloud: `create-react-app` is for development purposes only, `webpack` that is under the hood of it is not designed to serve files efficiently, and `nginx` is the simplest way to serve static files when you have them packaged with as little resources as possible.

So it matters in terms of size of the running container, and the memory consumption, this is why builder and multi-stage patterns are used in docker files

Dockerfile
EnormousCloud: +

EnormousCloud:
> I agree with @Siegrift that running the DAO dashboard locally is as simple as running `yarn && yarn start`. This allows for connecting to Ethereum mainnet out of the box, so I wonder how much of an issue this is 

To advocate it a bit, average user don't have the same version of Javascript, might not have `yarn` installed / prefer npm and have different experience of building . Earlier versions required to read a bit of source code (current situation is better). Webpack can consume 2gb of memory, so this limits its usage to the active JS developers only, and even if you know development process, you need to be highly insterested. Running docker with static files in the background has no consumption and is not noticeable.

Putting that on docker hub even removes a need of having `git` and turns the launch into a single command and 2-3 seconds of time - there will be no need to wait for the build

andreogle (approved):

Siegrift (approved):

bbenligiray:
Reminder to merge this?

PR #268, opened 2021-12-10
Add instructions to stop and remove container

dcroote:


Siegrift (approved):

README.md
```diff
@@ -18,6 +18,8 @@ docker run -d -p7770:80 --name api3-dao-dashboard api3-dao-dashboard
 
 This will create a API3 dashboard running on port 7770 of your localhost where it is safe to connect your wallet.
 
+Once you are finished interacting with the dashboard, the container can be stopped using `docker stop api3-dao-dashboard` and then removed using `docker rm api3-dao-dashboard`.
```
andreogle, line 21: Maybe we could also make a note that `docker run` could also be run with the flag `--rm` to avoid this? I forget all the time so this is my default.

https://docs.docker.com/engine/reference/run/#clean-up---rm

dcroote: Added- thanks for the suggestion.

PR #285, opened 2022-05-06
Fix HMR by pinning react-error-overlay to v6.0.9 (via a resolution)

mcoetzee:
I noticed that making any frontend code changes causes a runtime error in the HMR mechanism:
Screenshot 2022-05-06 at 14 08 28
https://github.com/facebook/create-react-app/issues/11773

Turned out to be an issue with react-scripts v4 (in combo with its react-error-overlay dependency). Long term fix would be to upgrade to react-scripts v5, but for now pinning react-error-overlay to v6.0.9 fixes the problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ yarn-error.log*
 
 # IDEs
 .vscode
+.idea
 
 # DAO contract deployments - Will be regenerated after every localhost deploy.
 # We still want deployments to mainnet/testnet checked in git.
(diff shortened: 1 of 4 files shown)


andreogle:
Nice catch. Just curious, is updating react-scripts to v5 a bit more than just bumping the version in package.json? I see some the breaking changes here: https://github.com/facebook/create-react-app/releases but haven't tried it myself

Siegrift (approved):
The new version of CRA uses Webpack 5, which does not include the polyfills it used to. This breaks the application due to packages not being found. I am super disappointed this is not mentioned in the changelog. This issue doesn't seem to have a good solution (apart from wasting time and trying to add the polyfills manually).

Anyway, the change LGTM 💪

PR #288, opened 2022-05-10
Change default proposal type to Secondary

### LINKED ISSUE #255: Proposal option radio button should select the Secondary option by default
Most proposals are of secondary type so this would prevent errors such as https://api3.eth.link/#/governance/primary-0

mcoetzee:
Fixes #255 

The cypress test has also been updated to assert that the default is Secondary.


Siegrift (approved):
👍 LGTM.

Nit: You can write the `Fixes #255` in a PR description instead of PR title

PR #290, opened 2022-05-25
DAO-152 Update Github actions cache keys (to force it to create new caches)

mcoetzee:
Currently updating the keys is the only way to do it: 
https://github.community/t/how-to-clear-cache-in-github-actions/129038
https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions


Siegrift:
Any idea how the cache got corrupted?

Siegrift (approved):

PR #291, opened 2022-06-01
DAO-153 ClaimsManager contract setup with initial integration

mcoetzee:
I wanted to load claims via the Claims Manager contract, which lead to adding the initial Policy Details and New Claim pages (in order to create the claims):

Screenshot 2022-06-01 at 16 45 28
Screenshot 2022-06-01 at 16 45 03

Logic around the claim statuses has also been updated to be in sync with the contract.

**Typechain:**
I chatted with Burak about the Typechain artefacts/types, and the claims-manager repo will be generating these artefacts for us, so for now I've manually generated and included just the ones we need (for ClaimsManagerWithKlerosArbitrator). These files will be removed at a later stage (when the claims-manager repo is ready).

**Contract Deployment:**
I also chatted with Burak about where the claims-manager repo should live. He created an isolated repo so that the auditors don't get flustered. So I investigated what it would take to include this claims-manager repo into the local development flow (i.e. the "prepare dao contracts" and "deploy dao contracts" steps). Turns out, with the Typechain stuff out of the way, it looks fairly easy to do it (I did a proof of concept). This PR leaves those dev-scripts as is for now, as it felt a bit premature to include that work for this PR


src/app.tsx
```diff
@@ -67,6 +69,12 @@ const AppContent = () => {
           <Route path="/claims" exact>
             <Claims />
           </Route>
+          <Route path="/policies/:policyId/claims/new" exact>
```
Siegrift, line 72: Currently, the proposals are not created this way (the form opens in a dialog when you are on proposals page). I fear it will feel inconsistent.

src/contracts/hooks.ts
```diff
@@ -9,6 +9,8 @@ import {
   Convenience__factory as ConvenienceFactory,
   TimelockManager__factory as TimelockManagerFactory,
 } from '../generated-contracts';
+// TODO DA0-151 Delete tmp folder and import from claims-manager package
```
Siegrift, line 12: It would be nicer if you created smaller subtasks for these.

src/logic/claims/data.ts
```diff
@@ -95,162 +85,93 @@ export function useUserClaimById(claimId: string) {
...
+    return { ids: [], byId: {} };
+  }
 
-      claim.deadline = calculateDeadline(claim);
-      acc[claim.claimId] = claim;
-      return acc;
-    }, {} as Record<string, Claim>);
-}
+  // Get all the dynamic data (status etc) via a single call
+  const calls = createdEvents.map((event) => {
```
Siegrift, line 104: Shouldn't we chunk this?

src/pages/claims/claim-list.tsx
```diff
@@ -1,6 +1,6 @@
-import { NavLink } from 'react-router-dom';
+import { Link } from 'react-router-dom';
```
Siegrift, line 1: Why change to `Link`? Shouldn't it be changed in other places as well?

src/pages/new-claim/new-claim.tsx
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  // Don't try to parse if the user is starting to type a negative number
+  if (form.amount.replace(/-/g, '').length === 0) {
+    messages.amount = 'Please fill in this field';
+  } else {
+    try {
+      const parsed = parseApi3(form.amount);
+      if (parsed.lte(0)) {
+        messages.amount = 'This must be greater than zero';
```
Siegrift, line 37: The violation message feels vague to me. WDYT?

```suggestion
        messages.amount = 'Amount must be greater than zero';
```

src/pages/new-claim/new-claim.tsx
```diff
@@ -0,0 +1,140 @@
...
+function getValidationMessages(form: FormState, policy: Policy) {
+  const messages: ValidationMessages<FormState> = {};
+
+  if (form.evidence.trim().length === 0) {
+    messages.evidence = 'Please fill in this field';
+  }
+
+  // Don't try to parse if the user is starting to type a negative number
+  if (form.amount.replace(/-/g, '').length === 0) {
+    messages.amount = 'Please fill in this field';
```
Siegrift, line 32: Nit: This message is a bit inaccurate.

I'd suggest using `goSync` and inside of it checking for the expected errors (positive number, <= coverage amount) and saying generic `Please enter a valid number` in case it fails.

This should wokr:
```
  const goAmountValidation = goSync(() => {
    const parsed = parseApi3(form.amount);

    if (parsed.lte(0)) return 'This must be greater than zero';
    if (parsed.gt(policy.coverageAmount)) return 'This must not exceed the coverage amount';
  });
  if (!goAmountValidation.success) messages.amount = 'Please enter a valid number';
  else if (goAmountValidation.data) messages.amount = goAmountValidation.data;

  return messages;
```

src/pages/claim-details/claim-actions.tsx
```diff
@@ -82,56 +121,48 @@ export default function ClaimActions(props: Props) {
...
-            <div className={styles.actionPanel}>
-              <Button type="secondary" disabled={disableActions} onClick={handleAppeal}>
-                Appeal
-              </Button>
-            </div>
+    case 'DisputeResolvedWithoutPayout':
+      return (
+        <div className={styles.actionSection}>
+          <p>Kleros</p>
+          <div className={styles.actionMainInfo}>Rejected</div>
```
Siegrift, line 153: Should this say `Rejected`? Shouldn't it be somethig like "resolved without payout"?

Siegrift:
> I chatted with Burak about the Typechain artefacts/types, and the claims-manager repo will be generating these artefacts for us, so for now I've manually generated and included just the ones we need (for ClaimsManagerWithKlerosArbitrator). These files will be removed at a later stage (when the claims-manager repo is ready).

I looked at the claims contract, but Typechain is not listed as dependency. I guess you have used it one time only and copied the artifacts to the DAO dashboard right?

> This PR leaves those dev-scripts as is for now, as it felt a bit premature to include that work for this PR

I would like this to be implemented as soon as possible as I would like to play with the app while reviewing. But I agree it should be done in a separate PR.

> I also chatted with Burak about where the claims-manager repo should live. He created an isolated repo so that the auditors don't get flustered

Do you know what is the final plan? Will the contract be moved to dao-contracts repository after audit?

src/pages/claims/claim-list.tsx
mcoetzee: I accidentally used NavLink previously, NavLink is for nav menu items (it has extra props for adding active classes etc)

src/pages/new-claim/new-claim.tsx
mcoetzee: Yeah sounds good

src/logic/claims/data.ts
mcoetzee: We don't expect users to create loads of claims (these are only the claims for the current user), so I would say probably not at this stage. Out of interest though, is it possible to load a subset of the events/logs?

mcoetzee:
> I looked at the claims contract, but Typechain is not listed as dependency. I guess you have used it one time only and copied the artifacts to the DAO dashboard right?

Correct ✅

> Do you know what is the final plan? Will the contract be moved to dao-contracts repository after audit?

At this stage, both Burak and I are happy to keep it as a separate repo. As mentioned, I did a proof of concept which worked out pretty chill, so unless you have strong feelings, or something comes up, I don't see the need to move it.

src/logic/claims/data.ts
Siegrift: >  Out of interest though, is it possible to load a subset of the events/logs?

Logs can be indexed and you can filter them by the indexed fields.

src/app.tsx
Siegrift: Is there any specific reason why not to include the claims to sidebar navigation?

Siegrift (approved):
I've tested also locally. 👍 LGTM

src/contracts/hooks.ts
```diff
@@ -78,6 +80,17 @@ export const useTimelockManager = () => {
...
 };
 
+export const useClaimsManager = () => {
+  const { provider, contracts } = useChainData();
+
+  return useMemo(() => {
+    if (!provider || !contracts) return null;
+
+    // TODO DAO-151 Remove REACT_APP_CLAIMS_MANAGER_ADDRESS and get the address from the "contracts" object
+    return ClaimsManagerFactory.connect(process.env.REACT_APP_CLAIMS_MANAGER_ADDRESS!, provider.getSigner());
```
Siegrift, line 90: Could you list this variable in `secrets.example.env` as well?

src/app.tsx
mcoetzee: We don't yet have the icon, but I can probably add it for testing convenience so long

src/logic/claims/data.ts
mcoetzee: Haha my question wasn't super clear (I know we can filter the logs as I do that already). I was just wondering hypothetically, if we were to chunk, whether it would apply to the logs. But no worries

Siegrift: I see. One way of doing that is specifying from/to which block search for the logs. But we can deal with that later...
E.g. we can just fetch all of them and once we see there are too many claims we can implement the chunking.

PR #294, opened 2022-06-04
Clarify main and production branch relationship

Siegrift:
Also updates the running from docker instructions to run DAO dashboard from "production" branch instead of "main".


mcoetzee (approved):
LGTM 👍

PR #296, opened 2022-06-08
DAO-156 Include claims-manager contracts in local development flow

mcoetzee:
### What does this change?
- Modifies the `prepare-dao-contracts-for-hardhat.js` `deploy-dao-contracts-on-hardhat.js` scripts to include the claims-manager repo
- Removes the `REACT_APP_CLAIMS_MANAGER_ADDRESS` usage and gets the contract address from contracts object

I modified the script to create a `dao-contract-repos` folder to clone both the api3-dao and claims-manager repo into:
```
dao-contract-repos/
  |- api3-dao/
  |- claims-manager/
```
One point of difference is that I used an alternative approach to install the claims-manager repo:
```
yarn add -D api3dao/claims-manager.git#0cf57283c1e21bad328b6aab78cc67614cad77fa
```
Which provides you with a cached tarball to download from Github. 

The alternative approach (`git+https://github.com/api3dao/claims-manager.git#somereference`) causes the yarn install process to regularly fail on build servers. Long story short, this approach clones the repo, and also installs the repo's dependencies (because there is a "prepare" script present in its `package.json`). This additional level of dependency installation is the problem (which you don't get when just downloading the tarball).


hardhat/hardhat.config.ts
```diff
@@ -66,12 +66,20 @@ task('create-user-policy', 'Creates a policy for the given user')
   .addParam('coverageAmount', 'The coverage amount')
   .addParam('ipfsHash', 'The IPFS policy hash')
   .setAction(async (args, hre) => {
+    const network = hre.network.name;
+    const deploymentFileName = `../src/contract-deployments/${network}-dao.json`;
```
mcoetzee, line 70: I followed the same approach here as the existing "sent-to-account" task (to handle the different networks). My question is though, should we be allowing these tasks to run on networks other than localhost? I can't imagine we would ever want to run this on `mainnet`

README.md
```diff
@@ -54,6 +54,13 @@ We use the `main` branch to develop new features. For production code, see the
 
 <!-- markdown-link-check-enable -->
 
+### Claims/Policies development
+
+You need a policy that is linked to your account in order to create a claim. You can create a policy for your account by
+running:
+
+`yarn create-user-policy --address <address> --coverage-amount 500 --ipfs-hash some-hash-001`
```
Siegrift, line 62: This doesn't work. It needs to be added to package.json (I see this is added in the other PR though).

Siegrift (approved):
👍 LGTM. Found minor bug

.gitignore
```diff
@@ -41,8 +41,8 @@ yarn-error.log*
 # Cypress blockchain configuration
 cypress-shared.json
 
-# Folder holding the DAO contracts repository used to deploy contracts locally for development and testing
-/dao-contracts
+# Folder holding the DAO contract repositories used to deploy contracts locally for development and testing
+/dao-contract-repos
```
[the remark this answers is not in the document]
mcoetzee: Yeah sure I'll keep it as `dao-contracts`

README.md
mcoetzee: I got a bit ahead of myself here, but ja, the other PR needs to be merged into main, and then I will also get rid of the last uses of REACT_APP_CLAIMS_MANAGER_ADDRESS (which was introduced with that branch) 👍

PR #297, opened 2022-06-22
DAO-157 Policy list page

mcoetzee:
### What does this change?
- Adds the Policies page with searching, filtering, and pagination
- Updates the New Claim page to capture an USD amount

- URL search params was used to drive the searching, filtering, and pagination
- I decided to write the pagination component myself, as the library options out there are unnecessarily clunky and not very composable
  - react-paginate: 22KB min, 5.4KB min+gz
  - rc-pagination: 16KB min, 5.2KB min+gz

### Screenshots
Screenshot 2022-06-22 at 17 23 18
Screenshot 2022-06-22 at 17 23 36
Screenshot 2022-06-22 at 17 25 14
Screenshot 2022-06-22 at 17 27 26
Screenshot 2022-06-22 at 17 28 09
Screenshot 2022-06-22 at 17 29 50

### DIFF
--- a/src/logic/policies/data.ts
+++ b/src/logic/policies/data.ts
@@ -117,5 +117,9 @@ async function loadPolicies(
 }
 
 export function isActive(policy: Policy) {
+  return isBefore(new Date(), policy.endTime);
+}
+
+export function canCreateClaim(policy: Policy) {
   return isBefore(new Date(), addDays(policy.endTime, 3));
 }
--- a/src/logic/policies/data.ts
+++ b/src/logic/policies/data.ts
@@ -1,9 +1,52 @@
-import { Policy, updateImmutablyCurried, useChainData } from '../../chain-data';
-import { useClaimsManager, usePossibleChainDataUpdate } from '../../contracts';
-import { useCallback, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
+import { addDays, isBefore } from 'date-fns';
 import { go } from '@api3/promise-utils';
+import { blockTimestampToDate, messages } from '../../utils';
+import { Policy, updateImmutablyCurried, useChainData } from '../../chain-data';
+import { ClaimsManagerWithKlerosArbitrator, useClaimsManager, usePossibleChainDataUpdate } from '../../contracts';
 import { notifications } from '../../components/notifications';
-import { blockTimestampToDate } from '../../utils';
+
+export function useUserPolicies() {
+  const { userAccount, policies, setChainData } = useChainData();
+  const claimsManager = useClaimsManager();
+  const sortedPolicies = useMemo(() => {
+    if (!policies.userPolicyIds) return null;
+    // Sort by policy end time in descending order
+    return policies.userPolicyIds
+      .map((policyId) => policies.byId![policyId]!)
+      .sort((a, b) => b.endTime.getTime() - a.endTime.getTime());
+  }, [policies]);
+
+  const [status, setStatus] = useState<'idle' | 'loading' | 'loaded' | 'failed'>('idle');
+  const loadUserPolicies = useCallback(async () => {
+    if (!claimsManager) return;
+
+    setStatus('loading');
+    const result = await go(() => loadPolicies(claimsManager, { userAccount }));
+
+    if (!result.success) {
+      notifications.error({ message: messages.FAILED_TO_LOAD_POLICIES, errorOrMessage: result.error });
+      setStatus('failed');
+      return;
+    }
+
+    setChainData(
+      'Loaded policies',
+      updateImmutablyCurried((state) => {
+        state.policies.userPolicyIds = result.data.ids;
+        state.policies.byId = { ...state.policies.byId, ...result.data.byId };
+      })
+    );
+    setStatus('loaded');
+  }, [claimsManager, userAccount, setChainData]);
+
+  usePossibleChainDataUpdate(loadUserPolicies);
+
+  return {
+    data: sortedPolicies,
+    status,
+  };
+}
 
 export function useUserPolicyById(policyId: string) {
   const { userAccount, policies, setChainData } = useChainData();
@@ -15,33 +58,10 @@ export function useUserPolicyById(policyId: string) {
     if (!claimsManager) return;
 
     setStatus('loading');
-    const result = await go(async () => {
-      const filter = claimsManager.filters.CreatedPolicy(null, userAccount, policyId);
-      const createdEvents = await claimsManager.queryFilter(filter);
-      const policiesById = createdEvents.reduce((acc, event) => {
-        const eventArgs = event.args;
-        const policy: Policy = {
-          policyId: eventArgs.policyHash,
-          claimant: eventArgs.claimant,
-          beneficiary: eventArgs.beneficiary,
-          coverageAmount: eventArgs.coverageAmount,
-          startTime: blockTimestampToDate(eventArgs.startTime),
-          endTime: blockTimestampToDate(eventArgs.endTime),
-          ipfsHash: eventArgs.policy,
-        };
-
-        acc[policy.policyId] = policy;
-        return acc;
-      }, {} as { [policyId: string]: Policy });
-
-      return {
-        policyIds: createdEvents.map((event) => event.args.policyHash),
-        policiesById,
-      };
-    });
+    const result = await go(() => loadPolicies(claimsManager, { userAccount, policyId }));
 
     if (!result.success) {
-      notifications.error({ message: 'Failed to load policies', errorOrMessage: result.error });
+      notifications.error({ message: messages.FAILED_TO_LOAD_POLICIES, errorOrMessage: result.error });
       setStatus('failed');
       return;
     }
@@ -49,7 +69,7 @@ export function useUserPolicyById(policyId: string) {
     setChainData(
       'Loaded policy',
       updateImmutablyCurried((state) => {
-        state.policies.byId = { ...state.policies.byId, ...result.data.policiesById };
+        state.policies.byId = { ...state.policies.byId, ...result.data.byId };
       })
     );
     setStatus('loaded');
@@ -61,4 +81,41 @@ export function useUserPolicyById(policyId: string) {
     data,
     status,
   };
+}
+
+async function loadPolicies(
+  contract: ClaimsManagerWithKlerosArbitrator,
+  params: { userAccount?: string; policyId?: string }
+): Promise<{ ids: string[]; byId: Record<string, Policy> }> {
+  const { userAccount = null, policyId = null } = params;
+  const createdEvents = await contract.queryFilter(contract.filters.CreatedPolicy(null, userAccount, policyId));
+
+  if (!createdEvents.length) {
+    return { ids: [], byId: {} };
+  }
+
+  const policiesById = createdEvents.reduce((acc, event) => {
+    const eventArgs = event.args;
+    const policy: Policy = {
+      policyId: eventArgs.policyHash,
+      claimant: eventArgs.claimant,
+      beneficiary: eventArgs.beneficiary,
+      coverageAmount: eventArgs.coverageAmount,
+      startTime: blockTimestampToDate(eventArgs.startTime),
+      endTime: blockTimestampToDate(eventArgs.endTime),
+      ipfsHash: eventArgs.policy,
+    };
+
+    acc[policy.policyId] = policy;
+    return acc;
+  }, {} as { [policyId: string]: Policy });
+
+  return {
+    ids: createdEvents.map((event) => event.args.policyHash),
+    byId: policiesById,
+  };
+}
+
+export function isActive(policy: Policy) {
+  return isBefore(new Date(), addDays(policy.endTime, 3));
 }
--- a/src/pages/new-claim/new-claim.tsx
+++ b/src/pages/new-claim/new-claim.tsx
@@ -1,14 +1,14 @@
 import { FormEventHandler, useState } from 'react';
 import { useParams } from 'react-router';
 import { goSync } from '@api3/promise-utils';
+import { BigNumber, utils } from 'ethers';
 import { BaseLayout } from '../../components/layout';
 import Input from '../../components/input';
 import Button from '../../components/button';
 import { isEmpty } from 'lodash';
+import { Policy } from '../../chain-data';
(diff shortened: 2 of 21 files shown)


src/pages/policies/policies.tsx
```diff
@@ -110,7 +110,7 @@ function PoliciesLayout(props: PoliciesLayoutProps) {
   const query = params.get('ipfs-hash') || '';
   const handleSubmit: FormEventHandler<HTMLFormElement> = (ev) => {
     ev.preventDefault();
-    const { value } = ev.currentTarget.query;
```
Siegrift: Are we sure `query` is going to be defined and reading `value` will not produce an error?

src/logic/policies/data.ts
```diff
@@ -117,5 +117,9 @@ async function loadPolicies(
 }
 
 export function isActive(policy: Policy) {
+  return isBefore(new Date(), policy.endTime);
```
Siegrift, line 120: Mhh, this is a change in logic. Is it correct?

src/pages/policies/policy-list.tsx
```diff
@@ -13,39 +15,60 @@ interface Props {
 }
 
 export default function PolicyList(props: Props) {
+  const history = useHistory();
```
Siegrift, line 18: nit: \n

(Someone should build a personal linter for my codestyle :p :D )

src/pages/new-claim/new-claim.tsx
```diff
@@ -21,18 +21,23 @@ interface FormState {
...
-function getValidationMessages(form: FormState) {
+function getValidationMessages(form: FormState, policy: Policy) {
   const messages: ValidationMessages<FormState> = {};
 
   if (form.evidence.trim().length === 0) {
     messages.evidence = 'Please fill in this field';
   }
 
-  const result = goSync(() => parseApi3(form.amount));
+  const result = goSync(() => parseClaimAmount(form.amount));
```
Siegrift, line 31: I see that the `parseClaimAmount` function also trims the value. Can you check if we should not do the same in the rest of the dashboard? (I think we handle validation quite well there, but just to make sure)

src/pages/policies/policies.module.scss
```diff
@@ -5,36 +5,34 @@
...
 }
 
 .clearButton {
   position: absolute;
-  right: 0;
+  right: 10px;
   svg {
-    width: 12px;
-    height: 12px;
+    width: 11px;
```
Siegrift, line 52: pixel perfect :p

src/pages/policies/policies.tsx
```diff
@@ -70,25 +86,69 @@ interface PoliciesLayoutProps {
 }
 
 function PoliciesLayout(props: PoliciesLayoutProps) {
+  const params = useQueryParams();
   const history = useHistory();
+
   const handleFilterChange = (showActive: boolean, showInactive: boolean) => {
+    const newParams = new URLSearchParams(params);
```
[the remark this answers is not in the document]
Siegrift: I also see that you've added it because you need more then 1 query param. In that case I agree that you're solution is more readable.

src/pages/policies/pagination.test.tsx
```diff
@@ -0,0 +1,40 @@
+import { paginate } from './pagination';
+
+describe('paginate()', () => {
```
Siegrift, line 3: 👍

src/pages/policies/pagination.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  it('accepts a custom page size', () => {
+    expect(paginate(42, { pageSize: 5, currentPage: 4 })).toEqual([1, 2, 3, 4, 5, '...', 8, 9]);
+    expect(paginate(42, { pageSize: 5, currentPage: 5 })).toEqual([1, '...', 4, 5, 6, '...', 9]);
+    expect(paginate(42, { pageSize: 5, currentPage: 6 })).toEqual([1, 2, '...', 5, 6, 7, 8, 9]);
+  });
+
+  it('handles when the current page is out of bounds', () => {
+    expect(paginate(200, { currentPage: -1 })).toEqual([1, 2, 3, 4, 5, '...', 19, 20]);
```
Siegrift, line 30: One thing that may be problematic is that if we want to support clicking on the `...` and doing some special behaviour. I don't think we care about it now, just giving a heads up.

src/pages/policies/pagination.tsx
```diff
@@ -73,60 +71,56 @@ export default function Pagination(props: Props) {
...
-    if (showDotsOnLeft && !showDotsOnRight) {
-      return [1, 2, '...', ...getNumberRange(lastPage - 4, lastPage)];
-    }
-
-    // 1 ... 5 6 7 ... 10
-    return [1, '...', ...getNumberRange(previousPage, nextPage), '...', lastPage];
-  }, [totalResults, pageSize, currentPage]);
+const DEFAULT_PAGE_SIZE = 10;
+
+export function paginate(
```
Siegrift, line 76: I'd like to use pagination in the governance section as well to make proposals loading more performant. For that we would need probably a bit more generic version, but I don't mind leaving that for a bit later.

I think for now most people use https://enormous.cloud/dao/api3/tracker/votings to browse past proposals, because on DAO dashboard it loads really slow.

src/pages/policies/pagination.tsx
```diff
@@ -73,60 +71,56 @@ export default function Pagination(props: Props) {
...
+  totalResults: number,
+  {
+    currentPage,
+    pageSize = DEFAULT_PAGE_SIZE,
+  }: {
```
Siegrift, line 81: Can you create a `PaginateOptions` type for this?

src/pages/policies/policies.tsx
```diff
@@ -48,7 +48,7 @@ export default function Policies() {
     return (
       <PoliciesLayout>
         <div className={styles.emptyState}>
-          <span>You need to be connected to view your policies.</span>
+          <span>You need to be connected to view your policies. </span>
```
Siegrift, line 51: Is the extra space needed? I'd prefer a padding instead.

src/pages/policies/policies.tsx
```diff
@@ -65,24 +65,51 @@ export default function Policies() {
...
-          <PolicyList policies={pagedPolicies} />
-          <Pagination totalResults={filteredPolicies.length} currentPage={currentPage} className={styles.pagination} />
-        </>
-      ) : policies.length === 0 ? (
-        <p className={styles.emptyState}>There are no policies linked to your account.</p>
-      ) : (
+  if (!filteredPolicies.length) {
+    let policyQualifier;
+    if (filter === 'active') {
+      policyQualifier = <span className={styles.highlight}>active </span>;
```
Siegrift, line 71: I'd prefer a CSS style instead of extra space.

src/pages/policies/policies.tsx
```diff
@@ -65,24 +65,51 @@ export default function Policies() {
...
+            <>
+              You don't have any policies associated with the connected address.
+              <br />
+              Connect an address associated with a policy to start a claim.
+            </>
+          ) : filter === 'none' ? (
+            <>Please select a filter.</>
+          ) : query ? (
+            <>
+              We couldn't find any {policyQualifier}policies with <span className={styles.highlight}>"{query}"</span>.
```
Siegrift, line 89: I guess you want to add a space character. `any {policyQualifier} policies..`

src/pages/policies/policies.tsx
```diff
@@ -65,24 +65,51 @@ export default function Policies() {
...
-      )}
+      </PoliciesLayout>
+    );
+  }
+
+  return (
+    <PoliciesLayout>
+      <div className={styles.resultsInfo}>
+        {filteredPolicies.length} result(s)
+        {query && <> for "{query}"</>}
```
Siegrift: Just for clarification: I am completely OK with a space here since you are building up a string, but I think it's better to use a CSS spacing for standalone texts (which I commented above).

src/pages/policies/policies.tsx
```diff
@@ -152,6 +175,10 @@ function PoliciesLayout(props: PoliciesLayoutProps) {
             block
           />
         </div>
+        <button type="submit" className={styles.searchButton}>
+          <SearchIcon aria-hidden />
+          <span className="sr-only">Submit</span>
```
Siegrift, line 176: I am not sure what `sr-only` means.  Can you be more specific?

Siegrift (commented):
So far I only looked at code. 

Will test the app and compare with the design tomorrow/the day after.

src/logic/policies/data.ts
mcoetzee: Hmm this is new logic for policies. Are you maybe thinking of claims?

src/pages/new-claim/new-claim.tsx
```diff
@@ -31,13 +31,8 @@ function getValidationMessages(form: FormState, policy: Policy) {
   const result = goSync(() => parseApi3(form.amount));
   if (!result.success) {
     messages.amount = 'Please enter a valid number';
-  } else {
-    const parsed = result.data;
-    if (parsed.lte(0)) {
-      messages.amount = 'Amount must be greater than zero';
-    } else if (parsed.gt(policy.coverageAmount)) {
```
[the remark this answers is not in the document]
mcoetzee: Hmm looks like you are commenting on a specific commit. That was subsequently reverted when it was clarified that both the coverage amount, and the claim amount is USD

src/pages/policies/policies.tsx
mcoetzee: This case is a bit different as it's not a block element, but a piece of inline text that gets highlighted, so adding the space will allow us to assert the text content of the parent element like "We couldn't find any active policies with ..." vs "We couldn't find any activepolicies with ..."

src/pages/policies/policies.tsx
mcoetzee: `policyQualifier` can be undefined (the space was added in the `policyQualifier` for convenience)

src/pages/policies/policies.tsx
mcoetzee: I copied it over from Tailwind: https://tailwindcss.com/docs/screen-readers#screen-reader-only-elements

And it stands for "screen reader only", so it's for making this button accessible by giving it text for screen readers, as the icon is for visual users only 👍 (a positive side effect is that this button becomes easily testable as well)

This util also exists for Bootstrap

src/pages/policies/policies.tsx
mcoetzee: Yeah the form event has all field elements available on it, and because this input has a name prop/attribute, it is available via `currentTarget.query` or `currentTarget.elements.query`.  Hashtag use the platform 🙂

src/pages/policies/pagination.test.tsx
mcoetzee: If we end up needing that, then it shouldn't require too much effort (the component could handle that)

src/pages/policies/pagination.tsx
mcoetzee: Re: your comment about the proposals, I would think the appropriate mechanism to drive paging would also be a query param, so this component would probably be generic enough already as it doesn't hardcode the path:
```tsx
return location.pathname + '?' + newParams.toString();
```

src/pages/policies/policies.tsx
```diff
@@ -70,25 +86,69 @@ interface PoliciesLayoutProps {
...
   };
 
-  const params = useQueryParams();
   const filter = params.get('filter');
   const activeChecked = !filter || filter === 'active';
   const inactiveChecked = !filter || filter === 'inactive';
+
+  const query = params.get('ipfs-hash') || '';
+  const handleSubmit: FormEventHandler<HTMLFormElement> = (ev) => {
+    ev.preventDefault();
```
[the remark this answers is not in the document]
mcoetzee: I think this gets a bit too subjective haha. Personally I don't think one can argue for or against newlines in specific places, because it's one of those things that is too fuzzy. I know a dev with close to 20 years under his belt, and he personally dislikes too many newlines, as it reduces the amount of code you can read without the need to scroll. 

But I wouldn't ever leave feedback for fuzzy things like this, it's in the eye of the beholder.

src/pages/new-claim/new-claim.tsx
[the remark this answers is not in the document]
Siegrift: I mean that we are no longer checking that the amount must be <= coverageAmount. But has that been reverted as well?

src/pages/policies/policies.tsx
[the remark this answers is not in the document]
Siegrift: Yeah, that's why I always consider them as "nits". Feel free to NOT implement them if you disagree.

src/logic/policies/data.ts
Siegrift: Mhh, from the diff I see that previously the `isActive` function was implemented as `isBefore(new Date(), addDays(policy.endTime, 3));` but now it's `isBefore(new Date(), policy.endTime)`

src/pages/new-claim/new-claim.tsx
Siegrift: Did you look at it? Do we handle the validation in the rest of the code correctly?

src/pages/policies/policies.tsx
Siegrift: I see.

src/pages/policies/policies.tsx
Siegrift: Sorry, missed that

Siegrift (approved):
👍 LGTM

src/logic/policies/data.ts
mcoetzee: Ah right (I see you were looking at a commit vs the end result). So a policy is considered inactive once it is past its end time, but you can still create a claim for another 72 hours after that (this logic moved to "canCreateClaim()").

src/pages/new-claim/new-claim.tsx
mcoetzee: I just dug a bit into this, and it seems we are missing some trimming for the New Proposal form and the Delegate form. E.g. valid addresses padded with spaces shows an "invalid address" validation message. I will create an issue for this 👍

PR #299, opened 2022-06-29
DAO-158 Multi-step create claim process

mcoetzee:
### What does this change?
- Adds `start-time` and `end-time` optional params to the `create-user-policy` Hardhat task
- Adds sequential steps:
    1. "Creating Evidence" (`instructions`)
    2. "Enter Claim Details" (`capture`)
    3. "Review Your Claim" (`confirmation`)

React state was used to control the steps as opposed to the URL, because you shouldn't be able to skip a step by typing in the corresponding URL, and you should be able to go back and edit the claim if you spot a mistake on the "Review Your Claim" step.

### Screenshots

Screenshot 2022-06-29 at 17 48 44
Screenshot 2022-06-29 at 17 47 55
Screenshot 2022-06-29 at 17 47 35
Screenshot 2022-06-29 at 17 48 25


src/pages/new-claim/claim-evidence-instructions.tsx
```diff
@@ -0,0 +1,57 @@
...
+          including on-chain evidence of affected transactions
+        </li>
+        <li>b) Show permanent damages or losses caused by the malfunction</li>
+        <li>c) Calculation method of claimed relief amount</li>
+      </ul>
+      <p>
+        Please also note that your claim information linked within the claim cannot be altered after submission, and all
+        supporting evidence presented must be pinned on IPFS and publicly available at the time of Claim submission
+        until the Claim proceedings are concluded and the coverage policy reaches its end time.{' '}
+        <a href="https://docs.google.com" target="_blank" rel="noopener noreferrer">
```
mcoetzee, line 24: I've added https://linear.app/api3-dao/issue/DAO-165 to update these links when finalised and to add add the finalised graphic/image to this step

src/pages/new-claim/new-claim-form.tsx
```diff
@@ -0,0 +1,114 @@
+import { FormEventHandler } from 'react';
+import { goSync } from '@api3/promise-utils';
+import { BigNumber, utils } from 'ethers';
+import isEmpty from 'lodash/isEmpty';
+import Input from '../../components/input';
+import Button from '../../components/button';
+import { Policy } from '../../chain-data';
+import globalStyles from '../../styles/global-styles.module.scss';
+import styles from './new-claim-form.module.scss';
```
mcoetzee, line 9: Locally this file was renamed from the previous `new-claim.tsx` file, but GH has decided to present it as a new file instead. The only logic that has changed from before is the `handleSubmit` function

.gitignore
```diff
@@ -17,6 +17,7 @@
 .env.development.local
 .env.test.local
 .env.production.local
+/tmp
```
Siegrift, line 20: Why put it inside .gitignore? Maybe we could generate this directly inside `generated-contracts` folder? Although this would probably require a structure change (to create subdirectories) so no need to do it now.

src/pages/new-claim/new-claim.tsx
```diff
@@ -0,0 +1,112 @@
...
+      setStatus('submitted');
+    } catch (err) {
+      setStatus('failed');
+    }
+  };
+
+  if (status === 'submitted') {
+    return (
+      <BaseLayout subtitle="New Claim">
+        <p>Success TODO</p>
```
Siegrift: I see some TODOs in the file - I guess this is because the labels are not yet finalized, right?

src/pages/new-claim/confirmation.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <ol className={styles.fieldList}>
+        <li>
+          <p>Enter the IPFS hash to your Claim Evidence form</p>
+          <p className={globalStyles.secondaryColor}>You created this hash in the previous step</p>
+          <p className={styles.confirmValue}>{form.evidence}</p>
+        </li>
+        <li>
+          <p>Requested relief amount, in USD</p>
+          <p className={globalStyles.secondaryColor}>
+            How much USD do you wish to receive? (Max of ${utils.commify(policy.coverageAmount.toString())})
```
Siegrift, line 29: I wonder, when is the user informed about the payment actually being in API3?

src/pages/new-claim/claim-evidence.tsx
```diff
@@ -0,0 +1,56 @@
+import Button from '../../components/button';
```
Siegrift: For some reason, this file breaks Github highlighting for me :)

src/pages/new-claim/claim-evidence.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export default function ClaimEvidence(props: Props) {
+  return (
+    <div className={styles.container}>
+      <p>Your claim must include the following:</p>
+      <ul className={styles.terms}>
+        <li>
+          a) Show a direct connection between the usage of your subscribed data feed and the claimed malfunction,
+          including on-chain evidence of affected transactions
```
Siegrift: What if the problem is a non existing data-feed update? In that case there is no transaction on chain that can be used as proof.

src/pages/new-claim/new-claim.tsx
```diff
@@ -86,7 +86,7 @@ export default function NewClaim() {
           <p className={globalStyles.bold}>Your claim ID is: {newClaimId}</p>
           <p className={styles.processMessage}>
             Your claim is being processed and will be voted on within 72 hours. Please check back for any updates and{' '}
-            <a href="https://docs.google.com" target="_blank" rel="noopener noreferrer">
+            <a href="https://docs.api3.org" target="_blank" rel="noopener noreferrer">
```
Siegrift, line 89: The document will be served from the api3.org page right?

Siegrift (commented):
Left some minor comments. I'll review once more after I come back from vacation.

.gitignore
mcoetzee: Oh this is in the project root dir. The `/tmp` root directory is something I got used to with Ruby on Rails. I sometimes use it to keep some temporary project related files

src/pages/new-claim/new-claim.tsx
mcoetzee: Hmm I'm not sure how, but you ended up commenting on a specific commit (and not the actual PR): https://github.com/api3dao/api3-dao-dashboard/pull/299/commits/81917d4dddb26acfd088bf5eb6f01a5767c9f804

I think this happened in the previous PR as well. Do you know how this happens as it's probably not where you meant to review?

src/pages/new-claim/confirmation.tsx
mcoetzee: There will be a claim process document that we point to on the "Thank you for submitting" section, which will document all the user needs to know. There is also a new requirement that we will show both the USD and API token amount on the Claims Details page when paid out (which should make it clear)

src/pages/new-claim/claim-evidence.tsx
mcoetzee: What is a non-existing data feed update? 🙂

src/pages/new-claim/new-claim.tsx
mcoetzee: The page doesn't exist yet, and it will be on docs.api3.org

mcoetzee: Yep I have a task already 👍

hardhat/hardhat.config.ts
```diff
@@ -59,6 +59,8 @@ task('create-user-policy', 'Creates a policy for the given user')
   .addParam('address', 'The user address')
   .addParam('coverageAmount', 'The coverage amount')
   .addParam('ipfsHash', 'The IPFS policy hash')
+  .addOptionalParam('startTime', 'The policy start time')
```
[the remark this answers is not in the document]
mcoetzee: Good catch. Will do 👍

Issue #106, opened 2021-06-16
Add e2e tests


Siegrift:
I'd like dashboard to be tested e2e as well...

Unfortunately, due to time pressure only simple tests were written, but I'd like to add more :) 

However, I think our tests (for such a simple page) should be under 15 minutes and we shouldn't complicate things and focus on the most importants parts and user flows.

## This includes

- [x] Initial setup = Done in https://github.com/api3dao/api3-dao-dashboard/pull/126
- [x] Test dashboard flow = Done in https://github.com/api3dao/api3-dao-dashboard/pull/126
- [x] Implement code changes allowing switching accounts in e2e tests (read description of https://github.com/api3dao/api3-dao-dashboard/pull/126)
- [x] #196
- [x] #157
- [ ] Test notifications (I think this can just be added to dashboard tests to save execution time)
- [x] #181
- [ ] Test active proposals page (voting, navigation to invalid proposal page, after epoch length passed proposal should move to history page)
- [ ] Test history proposals page (execute, invalid proposal details page)
- [ ] Delegation + treasury
- [ ] Test mobile version (probably just a bit to save execution time)
- [ ] We have a lot of annoying styling special cases. I think it's worth adding some visual tests as well (https://docs.cypress.io/guides/tooling/visual-testing#Functional-vs-visual-testing)

Siegrift:
There are many e2e tests for multiple user flows and there is little need for the ones in the issue checklist. Some points are even outdated (e.g. we have visual tests as well).

PR #300, opened 2022-07-04
DAO-167 Introduce visual testing with Percy

mcoetzee:
### What does this change?
- Adds 6 Percy snapshots to our existing Cypress tests
- Re-enables the New Proposal Cypress test (previously flaking test)
- Modifies the `login` command to close the error reporting notice if present
- Bumps `actions/upload-artifact` to v3 

Adding Percy is straight forward, but I had a long struggle figuring out why adding it to our Cypress run caused the tests to run super slowly on our GH action runner (often not finishing at all).

But long story short, having Cypress videos enabled on a GH action runner takes a large amount of the runner's CPU. So much that the video recordings don't actually work on main (you can download videos of old runs and see). This is a common problem (https://github.com/cypress-io/github-action/issues/483), so when I added the Percy process into the mix, it was the straw that broke the camel's back (ito CPU).

The cheap solution was to turn off video recording and to take screenshots of failures instead. This drastically reduces the runner's CPU usage and improves our current debugging experience (as the videos don't work).

### Github integration
There is a Percy GH integration we can install which I think makes a lot of sense: https://docs.percy.io/docs/github
We can configure it so that PR's that cause visual diffs can't be merged until those visual diffs have been reviewed/approved.
GitHub requires that you have the "owner" role in the organisation to install it, so @andreogle, you should be able to do that for the api3-dao-dashboard repo (as our current roles are set up).

### Percy snapshots

97d53acd0fb1bef2206007df8e78565e56bcb9bc18d06485579bfb7cb40a0ebe
b7a2683583111ddf27ba6fa8a9a18fa55285118237989bbf5eb3edf8e705e76f
01b5e9613478e206fc852571305dc956a5977475c52220d0f122cdf95b550841

121df54c4e25a373842363d653cc9748fdcead4bf178c133f979d39c6285fbf0
3a4a141eec6a0ca1270ccbb56c0212030bdeb75339d63b34c4b138fd259f0e2c
371a7d79e578066a476b4a096be8d2efacbb42ab882e9df78be06e06b8139309 (1)


Siegrift (commented):
I like this. I wanted to setup these visual tests at some point (see last item of https://github.com/api3dao/api3-dao-dashboard/issues/106#issue-922158061) but never had time to get to it. Nice to see it happen.

cypress/e2e/new-proposal.cy.ts
```diff
@@ -2,7 +2,7 @@
 
 import { ACCOUNTS, EPOCH_LENGTH } from '../support/common';
 
-it.skip('new proposal form validation', () => {
+it('new proposal form validation', () => {
```
andreogle, line 5: 👌🏻

.percy.json
```diff
@@ -0,0 +1,18 @@
+{
+  "version": 2,
+  "snapshot": {
+    "widths": [1280],
+    "minHeight": 1024,
+    "percyCSS": ".visual-test\\:invisible, .Toastify { visibility: hidden; }"
```
andreogle, line 6: So this is specifically to hide certain elements from Percy and more generally define any special CSS classes that Percy will apply?

src/components/timer/timer.tsx
```diff
@@ -93,7 +93,7 @@ const Timer = (props: Props) => {
           <>
             <div className={styles.timerColon}>:</div>
             <div className={styles.timerWrap}>
-              <div className={styles.timerNumber}>{timerSeconds}</div>
+              <div className={`visual-test:invisible ${styles.timerNumber}`}>{timerSeconds}</div>
```
andreogle, line 96: Why do we want to hide this?

andreogle (approved):

.percy.json
mcoetzee: Yep visual testing needs stable snapshots across runs, and sometimes you've got dynamic elements on the page which you need to hide for Percy to achieve a stable snapshot. Which is where this "percyCSS" comes into play. For HTML which is under your control you can use this `visual-test:invisible` util style (the `.Toastify` is to target the toaster library)

src/components/timer/timer.tsx
mcoetzee: The seconds counter is something that will probably cause snapshots to differ over runs, as a small delay in the test could cause the second to tick over and cause a visual diff

.percy.json
andreogle: Ah got it. That makes a lot more sense now, thanks 👍🏻

PR #301, opened 2022-07-05
DAO-166 Button component improvements

mcoetzee:
### What does this change?
- Uses a pure CSS solution for the Button underline
- Removes the wrapping div
- Renames the `type` prop to `variant` (type clashes with the HTML button type prop)
- Updates props to accept any HTML button prop

I used an absolutely positioned pseudo `::after` element to create the underline. In order to have no visual impact, a `margin-bottom` was added as well to the button to cater for the underline's space.

I went through all the Button usages where a `className` was used, and adjusted the CSS to no longer target a nested `button` element.

### DIFF
--- a/src/components/button/button.module.scss
+++ b/src/components/button/button.module.scss
@@ -1,8 +1,15 @@
 @import '../../styles/variables.module.scss';
 
-.buttonWrapper {
-  display: inline-block;
-  &.disabled {
+.button {
+  background-color: transparent;
+  position: relative;
+  color: $button-color;
+  border-width: 1px;
+  border-style: solid;
+  outline: none;
+  cursor: pointer;
+
+  &:disabled {
     opacity: 0.4;
     pointer-events: none;
   }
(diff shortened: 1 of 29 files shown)


src/components/button/button.module.scss
```diff
@@ -1,80 +1,85 @@
...
-    line-height: 22px;
-    padding: 20px 44px;
+.secondary {
+  border-color: $secondary-border;
 
-    @media (max-width: $max-md) {
-      font-size: $text-small;
-      padding: $space-md $space-xl;
-    }
+  &::after {
```
Siegrift, line 34: Don't you want `content: ''` here as well?

src/components/button/button.tsx
```diff
@@ -1,43 +1,28 @@
-import { ReactNode } from 'react';
+import { ComponentProps } from 'react';
```
Siegrift, line 1: This is nice! I didn't know about this. I've checked the `d.ts` file and it says to prefer `ComponentPropsWithoutRef` or `ComponentPropsWithRef` but I find them too verbose, so I'd even prefer if we just used the `ComponentProps`.

Siegrift (approved):
👍 LGTM.

I've checked the dashboard preview and run locally and didn't find any regressions.

src/components/button/button.tsx
mcoetzee: Yeah I am of the same opinion. Refs are such a rare thing to use, that I prefer the less verbose `ComponentProps` even if technically, that component can't yet receive a `ref`.

src/components/button/button.module.scss
mcoetzee: No need, all the common properties for primary and secondary are declared together here:
```scss
.primary,
.secondary {
  margin-bottom: 5px;

  &::after {
    content: '';
    height: 1px;
    width: calc(100% + 2px);
    position: absolute;
    bottom: -6px;
    left: -1px;
  }
}
```

PR #304, opened 2022-07-12
DAO-169 Use policy metadata instead of IPFS hash (to display and search)

mcoetzee:
### What does this change?
- Uses the new policy metadata field (a human-readable policy identifier) instead of the IPFS hash when searching and for displaying the policy list
- Updates the Hardhat task to accept metadata when creating a policy


src/pages/policies/policy-list.tsx
```diff
@@ -23,7 +23,7 @@ export default function PolicyList(props: Props) {
         <li key={policy.policyId} className={styles.policyItem}>
           <div className={styles.policyItemMain}>
             <Link className={styles.policyItemTitle} to={`/policies/${policy.policyId}`}>
-              Policy {policy.ipfsHash}
+              {policy.metadata}
```
Siegrift, line 26: Will there be some guarantee on metadata length? (Just asking whether because a long string might break/not-look-good in the UI)

hardhat/hardhat.config.ts
```diff
@@ -73,8 +75,8 @@ task('create-user-policy', 'Creates a policy for the given user')
       args.coverageAmount,
       BigNumber.from(Math.round(addDays(new Date(), -1).getTime() / 1000)),
       BigNumber.from(Math.round(addDays(new Date(), 30).getTime() / 1000)),
-      args.ipfsHash,
-      'Some policy metadata'
+      args.ipfsHash || 'Qm' + randomBytes(22).toString('hex'),
```
Siegrift, line 78: Nit: ```suggestion
      args.ipfsHash ?? 'Qm' + randomBytes(22).toString('hex'),
```

Siegrift (approved):
👍 LGTM. Left one small comment.

src/pages/policies/policy-list.tsx
mcoetzee: There is already CSS in place that truncates the text when it gets too long (like we do for proposals) 👍

Siegrift: Perfect. Thanks.

PR #312, opened 2022-07-28
DAO-163 Policy details page

mcoetzee:
### What does this change?
- Updates the existing policy page content
- Adds a "Back to X" link (we link to the policy page from both My Policies and My Claims)
- Removes the "Create a Claim" button from Policies page (it was decided in a design session that the user should first see the policy details before moving on to creating a claim)

### Screenshots
Screenshot 2022-07-27 at 17 55 54


src/pages/policy-details/policy-details.tsx
```diff
@@ -28,35 +32,73 @@ export default function PolicyDetails() {
   if (!policy) {
     return (
       <BaseLayout subtitle={`Policy ${policyId}`}>
-        <h4 className={styles.heading}>Policy {abbrStr(policyId)}</h4>
+        <h4 className={styles.heading}>Policy</h4>
```
Siegrift, line 38: We are not showing the policyId anymore, right? (Just double checking)

src/pages/policy-details/policy-details.tsx
```diff
@@ -28,35 +32,73 @@ export default function PolicyDetails() {
...
         {status === 'loading' && <p className={globalStyles.secondaryColor}>Loading...</p>}
         {status === 'loaded' && <p>Unable to find your policy with given id.</p>}
       </BaseLayout>
     );
   }
 
+  const policyIpfsHref = `https://ipfs.io/ipfs/${policy.ipfsHash}`;
```
Siegrift, line 45: We use the same logic (`getIpfsUrl`) in https://github.com/api3dao/api3-dao-dashboard/blob/main/src/pages/claim-details/claim-details.tsx#L49.

It would be nice to extract this to a helper function.

src/pages/policy-details/policy-details.tsx
```diff
@@ -28,35 +32,73 @@ export default function PolicyDetails() {
...
+              <p className={globalStyles.bold}>Beneficiary Address</p>
+              <p className={globalStyles.secondaryColor}>{policy.beneficiary}</p>
+            </div>
+            <div className={styles.detailsItem}>
+              <p className={globalStyles.bold}>Service Coverage Amount</p>
+              <p className={globalStyles.secondaryColor}>${formatUsd(policy.coverageAmountInUsd)}</p>
+            </div>
+            <div className={styles.detailsItem}>
+              <p className={globalStyles.bold}>Claims Allowed From</p>
+              <p className={globalStyles.secondaryColor}>{format(policy.claimsAllowedFrom, 'dd MMMM yyyy HH:mm')}</p>
```
Siegrift, line 102: Might be worth extracting the time format we use to some constant, so we are consistent about it in the project, wdyt?

src/pages/policy-details/policy-details.tsx
```diff
@@ -42,6 +45,19 @@ export default function PolicyDetails() {
   const policyIpfsHref = `https://ipfs.io/ipfs/${policy.ipfsHash}`;
   return (
     <BaseLayout subtitle={`Policy ${policyId}`}>
+      <div>
+        {params.get('from') === 'claims' ? (
```
Siegrift, line 49: Nice.

Siegrift (approved):
👍 LGTM.

Btw. do you know how long would it take to "revive" the preview deployments by deploying the contracts on some chain? I'd like to just click through a few buttons on the preview deployment instead of spinning up the local environment from scratch :)

src/pages/policy-details/policy-details.tsx
mcoetzee: 👍 We show the `policy.metadata`

Siegrift: Yeah, but in this field we show just a generic "Policy" so I meant that specifically.

src/pages/policy-details/policy-details.tsx
Siegrift: I think that is overkill. You could create the function like this:
```ts

const DATE_FORMATS = {
BASIC: "dd MM yyyy",
RICH: "dd MM yyyy hh:mm:ss"
}

function formatDate(date: Date, format: keyof typeof DATE_FORMATS) {
  return baseFormat(date, format);
}
```

I consider these similar to CSS constants and having those at one place always proved to be worth it. But I don't think this codebase is at the point where doing this would be necessary.

Siegrift:
> The deployment story is one that needs a lot clarification for me, ito how the new claims manager contracts will get deployed to all the different networks.

Won't it be deployed just to ETH mainnet? It needs to interact with the pool which only exists there.

> I also doubt whether we want to deploy unfinished contracts to a testnet?

Why not? (Apart from the fact that it's time consuming) You can just redeploy everything once there is an update.

PR #314, opened 2022-08-03
DAO-183 Fix policy and claim details loading (with new useChainUpdateEffect() hook)

mcoetzee:
### Problem statement
The existing `usePossibleChainDataUpdate()` hook accepts a callback that it triggers:
- on mount
- when account or network changes
- when a new block is mined

In my usage of this hook, my callback loads a policy by its id (which we get from the URL):
```ts
const callback = useCallback(async () => {
  // Load policy...
, [policyId]);

usePossibleChainDataUpdate(callback);
```
**The main issue:**
When you update the policy id in the URL, the callback doesn't get triggered again, and the UI gets out of sync.

**Additional issue:**
When multiple async calls are pending at the same time, there is no way to ensure that you only use the result from the latest call (as the async calls could resolve out of order).

### Proposed solution
Traditionally, these issues can be solved with `useEffect()`.

E.g. Basic example of loading data for a user (where user id can change at any time):
```ts
useEffect(() => {
  let isLatest = true;

  getData(userId).then(data => {
    if (isLatest) {
      setData(data);
    }
  });

  return {
    isLatest = false;
  }
}, [userId]);
```

So a solution would be to use a new hook that is similar to `useEffect()`, except that it also triggers:
- when account or network changes
- when a new block is mined

E.g. The same example as above, but with the new hook:
```ts
useChainUpdateEffect(() => {
  let isLatest = true;

  getData(userId).then(data => {
    if (isLatest) {
      setData(data);
    }
  });

  return {
    isLatest = false;
  }
}, [userId]);
```

Implementation details can be seen in this PR.

It's also worth noting that the `"react-hooks/exhaustive-deps"` Eslint rule can be extended to check custom hooks: 
```json
    "react-hooks/exhaustive-deps": [
      "warn",
      {
        "additionalHooks": "(useChainUpdateEffect)"
      }
    ]
```
Screenshot 2022-08-03 at 11 39 46

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,6 +2,12 @@
   "extends": ["react-app", "react-app/jest"],
   "rules": {
     "prefer-const": "error",
-    "no-console": ["error", { "allow": ["group", "groupEnd"] }]
+    "no-console": ["error", { "allow": ["group", "groupEnd"] }],
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      {
+        "additionalHooks": "(useChainUpdateEffect)"
+      }
+    ]
   }
 }
--- a/src/contracts/hooks.ts
+++ b/src/contracts/hooks.ts
@@ -183,6 +183,11 @@ export const usePossibleChainDataUpdate = (
   });
 };
 
+/**
+ * A hook that behaves like the useEffect hook, except that it also triggers the effect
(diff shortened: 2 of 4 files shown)


andreogle (approved):
Looks great to me 👍🏻

Siegrift (approved):
This looks great 👍

PR #316, opened 2022-08-12
DAO-186 Show remaining policy coverage on policy pages

mcoetzee:
### What does this change?
- It loads and displays the remaining coverage amount on the Policy Details and My Policies pages
- Adds a `upgrade-user-policy` hardhat task

~~The final design is still pending (it might end up being a progress bar), so this PR is about getting the plumbing in place. [DAO-187](https://linear.app/api3-dao/issue/DAO-187) has been logged to implement the final design.~~
**Edit:** With the upgradeable policies change https://github.com/api3dao/claims-manager/pull/40, it was raised by Burak that the remaining coverage amount is all we care about (and is what we upgrade), so we keep the design as is.

The My Policies page is paginated and the user can have many policies, so the remaining coverage gets loaded separately (from all the policies) and only for the current page of policies that the user is viewing.

The loading mechanism drives off of the current page's list of policies:
```tsx
export default function PolicyList(props: Props) {
  const policyIds = // derived from policies
  useRemainingCoverageLoader(policyIds);
```

On the Policy Details page, we load the remaining coverage at the same time as the policy, as their is no need to defer the loading.


src/pages/policies/policy-list.tsx
```diff
@@ -11,6 +11,9 @@ interface Props {
 }
 
 export default function PolicyList(props: Props) {
+  const policyIds = useStableIds(props.policies, (p) => p.policyId);
```
Siegrift, line 14: If I understand this correctly, we have this here to prevent unnecessary re-rerenders?

mcoetzee: To prevent unnecessary  loads ja. We trigger a load for the remaining coverage when the `policyIds` changes, so using this `useStableIds()` hook gives me a stable reference to those ids so that we don't trigger unnecessary loads (the reference to `props.policies` changes often as we refresh the data)

mcoetzee:
@Siegrift Changes have been made to ClaimsManager to make a policy upgradeable: https://github.com/api3dao/claims-manager/pull/40

Those changes impact this PR, so you can hold off further review until I've integrated those changes

Siegrift (approved):
👍 LGTM

PR #318, opened 2022-08-18
DAO-191 Remove counterOfferAmountInApi3 from claims

mcoetzee:
### What does this change?
- incorporates changes from https://github.com/api3dao/claims-manager/pull/44 i.e. the conversion of the settlement amount (USD -> API3) gets deferred to when the user accepts the settlement
- informing the user of what the converted API3 amount will be has been logged as a separate task [DAO-190](https://linear.app/api3-dao/issue/DAO-190)


Siegrift (approved):
👍 LGTM

PR #319, opened 2022-08-23
DAO-193 Fix proposal repeatedly loading on Proposal Details page

mcoetzee:
### What does this change?
It fixes the loop that the loading gets stuck in. The loop being:

1. The ID gets decoded into a BigNumber from the URL on render
2. The load gets triggered (as ID is a dependency)
3. The proposal finishes loading
4. Go to step 1


Siegrift (approved):
👍 Thanks for catching this.

PR #321, opened 2022-08-24
DAO-196 Sanitise URL in ExternalLink component

mcoetzee:
### What does this change?
Updates the ExternalLink component to only accept URLs that start with "https://" or "http://"

### DIFF
--- a/src/components/external-link/external-link.tsx
+++ b/src/components/external-link/external-link.tsx
@@ -9,7 +9,13 @@ interface Props {
 }
 
 const ExternalLink = (props: Props) => {
-  const { className, href, children } = props;
+  const { className, children } = props;
+
+  let href = props.href.trim();
+  const urlRegex = /^https:\/\//i; // Starts with https:// (case insensitive)
+  if (!urlRegex.test(href)) {
+    href = 'about:blank';
+  }
 
   return (
     <a href={href} className={classNames(className, styles.link)} target="_blank" rel="noopener noreferrer">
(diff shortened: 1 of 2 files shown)


src/components/external-link/external-link.tsx
```diff
@@ -9,7 +9,13 @@ interface Props {
 }
 
 const ExternalLink = (props: Props) => {
-  const { className, href, children } = props;
+  const { className, children } = props;
+
+  let href = props.href.trim();
+  const urlRegex = /^https:\/\//i; // Starts with https:// (case insensitive)
```
Siegrift, line 15: Mhh, we already do this precaution in https://github.com/api3dao/api3-dao-dashboard/blob/main/src/pages/proposal-commons/proposal-details/proposal-details.tsx#L39 but we also accept `http`

Any reason why only accept https?

mcoetzee: Just for some more context, this PR is more of a precautionary change (I created a proposal with a dodgy javascript URL and saw that it did not create a link for it). 

W.r.t. `http`, ah I missed the fact that we already allow them, so I will make the change to bring them back.

Siegrift: Also, it would be nice to show a warning (at least to console). WDYT?

mcoetzee: Sounds good 👍

Siegrift (approved):
👍 LGTM

PR #322, opened 2022-08-25
DAO-194 Fix false positive w.r.t. EVM script mismatch

mcoetzee:
### What does this change?
It fixes a false positive. The "This proposal is potentially malicious" warning incorrectly shows when the proposal has a non-zero ETH value. We currently don't have any proposals on mainnet with a non-zero ETH value, so no false positives are currently showing on mainnet.

Instead of providing the wei value to the `goEncodeEvmScript()` when re-encoding, the value gets formatted:

```diff
-  targetValue: decodedEvmScript.value.toString(),
+  targetValue: utils.formatEther(decodedEvmScript.value),
```

### DIFF
--- a/cypress/integration/malicious-proposal-validation.spec.ts
+++ b/cypress/integration/malicious-proposal-validation.spec.ts
@@ -34,7 +34,7 @@ describe('Malicious proposal validation', () => {
   it('does not show a warning if the original EVM script and the decoded EVM script matches up', () => {
     cy.increaseTimeAndRelogin(EPOCH_LENGTH + 60 * 60); // skip the genesis epoch (add 1 hour just to be sure)
     cy.exec(
-      `yarn create-proposal --type secondary --title 'API3 DAO BD-API Team Proposal' --description 'https://ipfs.fleek.co/ipfs/bafybeicfguu3bfhk3fyz5zn5wlujpksqxsaokse37674fhylouvsqnwd2m' --target-signature 'transfer(address,uint256)' --parameters '["0xCB943E4Fb0bCf7eC3C2E6D263c275b27F07701c6", "111363670000"]' --target-address 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 --target-value 0`
+      `yarn create-proposal --type secondary --title 'API3 DAO BD-API Team Proposal' --description 'https://ipfs.fleek.co/ipfs/bafybeicfguu3bfhk3fyz5zn5wlujpksqxsaokse37674fhylouvsqnwd2m' --target-signature 'transfer(address,uint256)' --parameters '["0xCB943E4Fb0bCf7eC3C2E6D263c275b27F07701c6", "111363670000"]' --target-address 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 --target-value 10`
     );
 
     cy.findAllByText('Governance').filter(':visible').click();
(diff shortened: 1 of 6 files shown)


bbenligiray:
I think the problem here is that the dashboard keeps the value in ether and not Wei

mcoetzee:
> I think the problem here is that the dashboard keeps the value in ether and not Wei

@bbenligiray Hmm the `goEncodeEvmScript()` accepts the form data as input, so it is understandable to me why it expects the ether amount as it is what the user types in. Are you thinking that it should rather expect a BigNumber (the wei)?

bbenligiray:
The form is meant to specify an arbitrary transaction in a very literal way and the `value` field of a transaction is in Wei. So the form expects the value field to be in Wei.

mcoetzee:
> The form is meant to specify an arbitrary transaction in a very literal way and the `value` field of a transaction is in Wei. So the form expects the value field to be in Wei.

Oh I get it now. So a better solution would be to ask the user to input in Wei rather than Ether. Shall I go ahead and make that change?

bbenligiray:
Yeah

Siegrift:
Yeah, we also mention [here](https://github.com/api3dao/api3-dao-dashboard/blob/main/src/pages/proposals/forms/new-proposal-form.tsx#L152) that we expect the number to be in WEI.

Siegrift:
> I just saw that, which makes me wonder whether the field label even needs change, as it seems like this is what is was supposed to be...

This is just the tooltip, so it's shown only on hover. I think it's a good idea to have add "in wei" to the value label.

bbenligiray:
If there is a tooltip no need to change the label imo. I'm 100% this was discussed already.

Siegrift (approved):
👍 LGTM.

Btw. we show the value in proposal details [formatted in ETH](https://github.com/api3dao/api3-dao-dashboard/blob/format-target-value/src/pages/proposal-commons/proposal-details/proposal-details.tsx#L250), but I think that is OK.

mcoetzee:
> 👍 LGTM.
> 
> Btw. we show the value in proposal details [formatted in ETH](https://github.com/api3dao/api3-dao-dashboard/blob/format-target-value/src/pages/proposal-commons/proposal-details/proposal-details.tsx#L250), but I think that is OK.

Yeah that is one thing that gives me pause (wrt keeping the labels as is):
Screenshot 2022-08-25 at 17 00 24
Screenshot 2022-08-25 at 17 00 48

Another thing that came up in my testing just now is decimals. A.t.m. the encoding function produces a "Please enter a valid non-negative amount in Wei" validation message when they attempt to enter decimals. Are we happy with that? FYI the number input library we use is probably able to prevent the user from entering decimals

Siegrift:
> Yeah that is one thing that gives me pause (wrt keeping the labels as is):

Ok, now that I see that the label is called the same, but the value has different denominations I no longer think that "this is OK". If Burak wants the label to stay the same I think we should show the raw amount in proposal details as well. WDYT?

mcoetzee:
> Ok, now that I see that the label is called the same, but the value has different denominations I no longer think that "this is OK". If Burak wants the label to stay the same I think we should show the raw amount in proposal details as well. WDYT?

That sounds sensible to me 👍

mcoetzee:
**In summary**

We show the raw amount on the Proposal Details page:
Screenshot 2022-08-25 at 17 00 24

Screenshot 2022-08-26 at 10 39 54

And if you attempt to enter any decimal value, then it triggers a validation error:
Screenshot 2022-08-26 at 10 40 57

Siegrift:
👍 LGTM

bbenligiray (approved):

PR #323, opened 2022-08-29
DAO-197 Dependency housekeeping 

mcoetzee:
### What does this change?
- sorts out nested dependencies
  - ansi-regex
  - async
  - json-schema
  - path-parse
- removes `npx` as it is bundled with `npm`, which in turn is bundled with node
- replaces the deprecated `node-sass` with `sass`

These changes reduce the size of `node_modules` from 1GB down to 720MB

~~I deleted and regenerated the `yarn.lock` file.~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,6 @@
   },
   "resolutions": {
     "@types/jest": "^26.0.24",
-    "@types/react": "^17.0.39",
     "react-error-overlay": "6.0.9"
   }
 }
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "@testing-library/user-event": "^13.5.0",
     "@typechain/ethers-v5": "^8.0.2",
     "@types/chai": "^4.3.0",
-    "@types/jest": "^27.4.0",
+    "@types/jest": "^26.0.24",
     "@types/lodash": "^4.14.178",
     "@types/mocha": "^9.1.0",
     "@types/node": "^17.0.15",
@@ -123,9 +123,8 @@
     "typescript-plugin-css-modules": "^3.4.0"
   },
   "resolutions": {
-    "//": "Remove once upgraded to react-scripts v5. See https://github.com/facebook/create-react-app/issues/11773",
+    "@types/jest": "^26.0.24",
     "react-error-overlay": "6.0.9",
-    "//": "Remove once upgraded to react v18. react-router and react-input-autosize cause @types/react to install v18",
     "@types/react": "^17.0.39"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -112,11 +112,10 @@
     "hardhat": "^2.8.3",
     "husky": "^7.0.4",
     "lint-staged": "^12.3.3",
-    "node-sass": "^7.0.1",
-    "npx": "^10.2.2",
```
Siegrift, line 116: I must have installed this by mistake. It makes no sense to have this in deps.

Siegrift:
> I deleted and regenerated the yarn.lock file.

You can use `yarn upgrade-interactive --latest` which lets you pick updates interactively (e.g. do just dev dependencies, just patch+minor, etc..)

mcoetzee:
Part of the problem is that the deps we're looking to sort out are deeply nested deps like `async`:
```
react-scripts > webpack-dev-server > portfinder > async
```
We are already on the latest patch of react-scripts v4 (4.0.3), so the `yarn upgrade-interactive` doesn't give you the option to upgrade `react-scripts`.

And the `yarn.lock` does look a little sketchy to me on main (e.g. it's got 3 versions of @babel/core installed), so my thinking was to see how it goes by simply regenerating it. 

I did pick up an issue with the Fleek build though, which can be resolved with another resolution.

mcoetzee:
I ended up taking a less drastic approach (instead of regenerating the lock file), because I spotted too big of an increase in the bundle size. 

Reinstalling the below deps got us the desired result:
```
@walletconnect/web3-provider
@aragon/os
@testing-library/cypress
@testing-library/jest-dom
@testing-library/react
@types/jest
concurrently
cypress 
hardhat 
lint-staged
react-scripts
replace-in-file
```

package.json
```diff
@@ -94,7 +94,7 @@
     "@testing-library/user-event": "^13.5.0",
     "@typechain/ethers-v5": "^8.0.2",
     "@types/chai": "^4.3.0",
-    "@types/jest": "^27.4.0",
+    "@types/jest": "^26.0.24",
```
mcoetzee, line 97: `react-scripts` v4 is on v26 of jest, so bumped this down to match it

package.json
```diff
@@ -112,19 +112,18 @@
     "hardhat": "^2.8.3",
     "husky": "^7.0.4",
     "lint-staged": "^12.3.3",
-    "node-sass": "^7.0.1",
-    "npx": "^10.2.2",
     "prettier": "^2.5.1",
-    "react-scripts": "4.0.3",
+    "react-scripts": "^4.0.3",
```
Siegrift, line 116: The convention for this is to stay pinned, but it doesn't matter much.

Siegrift (approved):
👍 LGTM

PR #326, opened 2022-09-06
DAO-189 Integrate claim hash, dispute ID, and date type changes

mcoetzee:
### What does this change?

- Integrates changes from https://github.com/api3dao/claims-manager/pull/50 (`claimIndex` -> `claimHash`)
- Integrates changes from https://github.com/api3dao/claims-manager/pull/53
- Integrates changes from https://github.com/api3dao/claims-manager/pull/60 (use `number` instead of `BigNumber` for dates)
- Includes dispute with ID 0 again. https://github.com/api3dao/claims-manager/issues/43 was fixed


src/logic/claims/data.ts
```diff
@@ -132,20 +129,16 @@ async function loadClaims(
...
-      disputeEvents.filter((ev) => ev.args.disputeId.gt(0))
-    ),
+    getDisputeContractData(arbitratorProxy, disputeEvents),
   ]);
 
   // Combine the static and dynamic data
   const claimsById = createdEvents.reduce((acc, event, index) => {
     const eventArgs = event.args;
-    const claimId = eventArgs.claimIndex.toString();
+    const claimId = eventArgs.claimHash;
```
Siegrift, line 138: Should we call it claimHash in the app as well? Probably not worth the hassle right?

Siegrift (approved):
👍 LGTM

src/logic/claims/data.ts
mcoetzee: Yeah I also don't think it's worth it 👍

PR #327, opened 2022-09-07
DAO-203 Create pages/components folder for domain specific building blocks

mcoetzee:
### What does this change?
- Moves `src/components/policies` -> `pages/components/policies`
- Moves `pages/proposal-commons` contents into `pages/components` folder


Siegrift (approved):
👍 Thanks. 

I really like the separation is now implemented properly.

PR #330, opened 2022-09-12
DAO-207 Prevent overwriting an existing localhost-dao.json file in postinstall script

mcoetzee:
### What does this change?
It prevents overwriting an existing `localhost-dao.json` file in the postinstall script, so that you don't lose your deployed addresses unnecessarily.

If there is an existing `localhost-dao.json` file, we only continue overwriting it if its keys (i.e. contract names) don't match the example file's keys.


Siegrift (approved):
👍 LGTM

PR #332, opened 2022-09-14
DAO-202 Integrate policy downgrades

mcoetzee:
### What does this change?
- pulls in latest changes from https://github.com/api3dao/claims-manager/pull/65 
- adds a `downgrade-user-policy` Hardhat task
- includes the `DowngradedPolicy` log in the policy loading strategy

We combine the `UpgradedPolicy` and `DowngradedPolicy` events, sort them, and use the last (most recent) event in order to get the latest `claimsAllowedUntil` value.


hardhat/hardhat.config.ts
```diff
@@ -132,8 +131,8 @@ task('create-user-policy', 'Creates a policy for the given user')
       userAddress,
       userAddress,
       parseUsd(args.coverageAmount),
-      BigNumber.from(Math.round(claimsAllowedFrom.getTime() / 1000)),
-      BigNumber.from(Math.round(claimsAllowedUntil.getTime() / 1000)),
+      Math.round(claimsAllowedFrom.getTime() / 1000),
```
Siegrift, line 134: I guess this was just to reduce characters, right?

src/logic/policies/data.ts
```diff
@@ -124,27 +124,30 @@ export function useUserPolicyById(policyId: string) {
...
     contract.queryFilter(contract.filters.CreatedPolicy(null, userAccount, policyId)),
     contract.queryFilter(contract.filters.UpgradedPolicy(null, userAccount, policyId)),
+    contract.queryFilter(contract.filters.DowngradedPolicy(null, userAccount, policyId)),
   ]);
 
   if (!createdEvents.length) {
     return { ids: [], byId: {} };
   }
 
+  const stateChangedEvents = sortEvents([...upgradedEvents, ...downgradedEvents] as const);
```
Siegrift, line 137: Nit: Do you need as const here?

src/utils/logs.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    expect(result).toEqual([
+      { testId: 'A', blockNumber: 0, logIndex: 0 },
+      { testId: 'B', blockNumber: 1, logIndex: 0 },
+      { testId: 'C', blockNumber: 2, logIndex: 1 },
+      { testId: 'D', blockNumber: 2, logIndex: 2 },
+      { testId: 'E', blockNumber: 2, logIndex: 3 },
+      { testId: 'F', blockNumber: 3, logIndex: 0 },
+      { testId: 'G', blockNumber: 4, logIndex: 0 },
+    ]);
+    expect(result === events).toBe(false); // Sanity check. The function should return a new array
```
Siegrift, line 26: Nit:
```suggestion
    expect(result === events).toBe(false); // sortEvents is immutable
```

might be more descriprive.

Siegrift (approved):
👍 LGTM

src/logic/policies/data.ts
mcoetzee: Yeah this one was unexpected. Without `as const` TS incorrectly infers `stateChangedEvents` as `UpgradedPolicyEvent[]`. With `as const` TS correctly infers it as `UpgradedPolicyEvent | DowngradedPolicyEvent)[]`.

hardhat/hardhat.config.ts
mcoetzee: Yeah it accepts `BigNumberish` arguments here, to the `BigNumber.from(...)` is unnecessary

src/logic/policies/data.ts
Siegrift: Oh, I see. That seems one of the many bugs they have.

PR #333, opened 2022-09-16
DAO-176 Show Escalate and Appeal confirmation modals

mcoetzee:
### What does this change?
- shows an Escalate and Appeal confirmation modal on the Claim Details page when the user decides to escalate or appeal
- adds a reusable Skeleton component (used as the loading state when loading the escalate/appeal cost)

The skeleton styling is consistent with https://market.api3.org/dapis, and sourced from https://github.com/dvtng/react-loading-skeleton/blob/master/src/skeleton.css

### Screenshots
Screenshot 2022-09-16 at 16 40 05

https://user-images.githubusercontent.com/747979/190662561-02664c01-d3da-4c0d-92ae-053d67fa8f7a.mov


Siegrift (approved):
Code 👍 LGTM

PR #334, opened 2022-09-20
DAO-208 Handle updates to Policy metadata

mcoetzee:
### What does this change?
- extends policy data loading to accommodate changes from https://github.com/api3dao/claims-manager/pull/69
- adds a `update-policy-metadata` Hardhat task
- extends claims data loading to include policy metadata (the metadata will be used in the claims list [DAO-159](https://linear.app/api3-dao/issue/DAO-159)) 
- integrates changes from https://github.com/api3dao/claims-manager/pull/74

- the new `AnnouncedPolicyMetadata` event gets used to load the policies' metadata
- the `Claim` type was extended with a new field:
```ts
  policy: { id: string; metadata: string };
```


hardhat/hardhat.config.ts
```diff
@@ -127,15 +127,25 @@ task('create-user-policy', 'Creates a policy for the given user')
     const claimsAllowedUntil = args.claimsAllowedUntil
       ? parseISO(args.claimsAllowedUntil)
       : addDays(claimsAllowedFrom, 30);
-    const tx = await claimsManager.createPolicy(
+    const ipfsHash = args.ipfsHash || 'Qm' + randomBytes(22).toString('hex');
+
+    // We use a multicall in order to create the policy and announce its metadata in the same transaction
```
Siegrift, line 132: Do you need to do that using mutlicall?

mcoetzee: Yeah that's how it will be called from the API3 market, because we want the `AnnouncedPolicyMetadata` event in the same block/transaction as the `CreatedPolicy` event.

Siegrift (approved):
👍 LGTM

PR #335, opened 2022-09-21
DAO-181 Add policy & claim icons and External Links section to navigation

mcoetzee:
### What does this change?
- uses new icons in nav menu
- adds External Links section to nav menu
- renames pages/claims -> pages/my-claims

### Screenshots
Screenshot 2022-09-21 at 14 45 56

Screenshot 2022-09-21 at 14 56 36


src/pages/components/treasury/treasury.tsx
```diff
@@ -46,13 +47,13 @@ const TreasuryDropdown = (props: TreasuryDropdownProps) => {
           <span className={classNames(styles.copy, globalStyles.textSmall)}>
             <Tooltip overlay={etherscanExplainer}>
               <ExternalLink href={getEtherscanAddressUrl(chainId, agentAddress) ?? ''} className={styles.dropdownLink}>
-                <img
-                  src={images.externalLink}
-                  alt={etherscanExplainer}
+                <ExternalLinkIcon
+                  aria-hidden
```
Siegrift, line 51: btw. how are we doing with the accessibility of the site? Have you checked that or you are just setting these attributes as a best practice? 

I didn't care about that early on, but since you've added this for multiple components (especially for the new stuff) it might be nice to do it properly everywhere.

Siegrift (approved):
👍 LGTM. Love the icons :)

src/pages/components/treasury/treasury.tsx
mcoetzee: Yeah wrt the `aria-hidden` stuff, icons are only meaningful for visual users (so it's best practice to tell the screen reader to ignore them), and to provide text that the visual users can't see (if there isn't already text that gives meaning).

But in general, I don't think the app is doing too bad ito a11y. Perhaps when the dust settles, I can do a bit of an a11y audit.

PR #336, opened 2022-09-27
DAO-209 Load and display payout info on the Claim Details page

mcoetzee:
### What does this change?
- it loads payout data on the Claim Details page
- it displays the payout amount (API3 tokens), a help/info tooltip, and the transaction link for claims with status:
    - `ClaimAccepted`
    - `SettlementAccepted`
    - `DisputeResolvedWithClaimPayout`
    - `DisputeResolvedWithSettlementPayout`
- it preloads payout data on the My Claims page in order to improve the Claims Details page load

- the `AcceptedClaim`, `AcceptedSettlement`, `ResolvedDisputeByAcceptingClaim`, `ResolvedDisputeByAcceptingSettlement` events were used to load the payout data for the claims
- to simplify the Claim Details page, the claim and the payout data is loaded together

### Screenshots
Screenshot 2022-09-26 at 15 44 06

### DIFF
--- a/src/contracts/network.ts
+++ b/src/contracts/network.ts
@@ -37,9 +37,7 @@ export const ETHERSCAN_HOSTS: { [chainId: string]: string } = {
 };
 
 export const getEtherscanTransactionUrl = (chainId: number | undefined, transactionHash: string) => {
-  if (!chainId) return;
-
-  const host = ETHERSCAN_HOSTS[chainId];
+  const host = chainId ? ETHERSCAN_HOSTS[chainId] : null;
   if (!host) return;
 
   // For example: https://ropsten.etherscan.io/tx/0xe4394ea70b32486f59f92c5194c9083bd36c99f2f0c32cfc9bacce3486055d24
@@ -47,9 +45,7 @@ export const getEtherscanTransactionUrl = (chainId: number | undefined, transact
 };
 
 export const getEtherscanAddressUrl = (chainId: number | undefined, address: string) => {
-  if (!chainId) return;
-
-  const host = ETHERSCAN_HOSTS[chainId];
+  const host = chainId ? ETHERSCAN_HOSTS[chainId] : null;
   if (!host) return;
 
   // For example: https://etherscan.io/address/0xb0A20975f540656E331e2331C6caEc608Ff254fc
--- a/src/logic/claims/data.ts
+++ b/src/logic/claims/data.ts
@@ -395,28 +395,28 @@ async function loadClaimPayoutData(contract: ClaimsManager, params: { userAccoun
   const [acceptedEvents, settlementEvents, resolvedClaimEvents, resolvedSettlementEvents] = await Promise.all([
     contract.queryFilter(
       contract.filters.AcceptedClaim(
-        // @ts-expect-error
+        // @ts-expect-error Typechain doesn't recognise that you can provide an array for any filter topic
         claimIds,
         userAccount
       )
     ),
     contract.queryFilter(
       contract.filters.AcceptedSettlement(
-        // @ts-expect-error
+        // @ts-expect-error Typechain doesn't recognise that you can provide an array for any filter topic
         claimIds,
         userAccount
       )
     ),
     contract.queryFilter(
       contract.filters.ResolvedDisputeByAcceptingClaim(
-        // @ts-expect-error
+        // @ts-expect-error Typechain doesn't recognise that you can provide an array for any filter topic
         claimIds,
         userAccount
       )
     ),
     contract.queryFilter(
       contract.filters.ResolvedDisputeByAcceptingSettlement(
-        // @ts-expect-error
+        // @ts-expect-error Typechain doesn't recognise that you can provide an array for any filter topic
         claimIds,
         userAccount
       )


src/contracts/network.ts
```diff
@@ -37,12 +36,14 @@ export const ETHERSCAN_HOSTS: { [chainId: string]: string } = {
   42: 'https://kovan.etherscan.io',
 };
 
-export const getEtherscanTransactionUrl = (transaction: ethers.Transaction) => {
-  const host = ETHERSCAN_HOSTS[transaction.chainId.toString()];
+export const getEtherscanTransactionUrl = (chainId: number | undefined, transactionHash: string) => {
+  if (!chainId) return;
```
Siegrift, line 40: Nit: Maybe combine these two ifs?

src/logic/claims/data.ts
```diff
@@ -98,7 +109,8 @@ export function useUserClaimById(claimId: string) {
       setChainData(
         'Loaded claim',
         updateImmutablyCurried((state) => {
```
Siegrift, line 111: Retrospectively, I really dislike this `updateImmutablyCurried` name. 
Maybe at this point we might plug in some proper state management library like zustand/redux to help with performance.

WDYT?

src/logic/claims/data.ts
```diff
@@ -329,3 +361,64 @@ function getDisputeStatus(period: DisputePeriod): DisputeStatus {
...
+
+  const events = sortEvents([
+    ...acceptedEvents,
+    ...settlementEvents,
+    ...resolvedClaimEvents,
+    ...resolvedSettlementEvents,
+  ] as const);
+
+  const byId = events.reduce((acc, ev) => {
+    if ('clippedAmountInUsd' in ev.args) {
```
Siegrift, line 435: You probably want to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

src/logic/claims/data.ts
```diff
@@ -329,3 +361,64 @@ function getDisputeStatus(period: DisputePeriod): DisputeStatus {
       return 'Solved';
   }
 }
+
+async function loadClaimPayoutData(contract: ClaimsManager, params: { userAccount?: string; claimIds?: string[] }) {
+  const { userAccount = null, claimIds = null } = params;
+  const [acceptedEvents, settlementEvents, resolvedClaimEvents, resolvedSettlementEvents] = await Promise.all([
+    contract.queryFilter(
+      contract.filters.AcceptedClaim(
+        // @ts-expect-error
```
Siegrift, line 398: nit: provide reason

src/pages/claim-details/claim-actions.tsx
```diff
@@ -180,7 +180,17 @@ export default function ClaimActions(props: Props) {
               </Tooltip>
             </div>
           </div>
-          <p className={styles.actionMessage}>All done! The claim payout has been accepted.</p>
+          <p className={styles.actionMessage}>
+            All done! The claim payout has been accepted.{' '}
+            <a
```
Siegrift: Maybe use the `ExternalLink` compoent?

mcoetzee: Yeah that's the plan 👍. I've logged a separate issue to make ExternalLink easier to reuse (it's got unnecessary styling that makes it a bit of a pain): https://linear.app/api3-dao/issue/DAO-211/remove-styling-from-externallink-component-to-make-it-easier-to-reuse

As part of that task I will use ExternalLink where appropriate

src/logic/claims/data.ts
mcoetzee: I'm not a fan of the `updateImmutablyCurried` name either, but I don't think we need to reach for a state management library. The app is simple enough. We can probably get rid of `updateImmutablyCurried` though and tweak the `setChainData()` API to:
```ts
setChainData(
  'Loaded policies',
  (state) => {
    state.policies.userPolicyIds = result.data.ids;
    state.policies.byId = { ...state.policies.byId, ...result.data.byId };
  }
);
```

Siegrift: or simply call it `updateState` ?

src/logic/claims/data.ts
mcoetzee: Using `in` is a common way to narrow down types in Typescript: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing

Siegrift: Yeah, but I dislike the fact that `in` traverses the full prototype chain and in some cases results in wrong logic.

mcoetzee: I think this is a matter of preference 🙂. I highly doubt we are at risk of any wrong logic here ito inherited properties.
Also, using `hasOwnProperty()` doesn't give us the type narrowing we want here:
Screenshot 2022-09-29 at 12 48 23

Lastly, there is also the Eslint rule against using `hasOwnProperty()` directly: https://eslint.org/docs/latest/rules/no-prototype-builtins

src/logic/claims/data.ts
mcoetzee: Hmm Immer uses `produce`, so I reckon we should probably follow their lead and call it `produceState`. My feeling is `setChainData` and `updateState` sounds too similar to each other.

src/logic/claims/data.ts
Siegrift: Yeah, feel free to leave it as is. Especially the narrowing is a big plus.

src/logic/claims/data.ts
Siegrift: I am fine with `produceState` as well.

mcoetzee: FYI I will rename in https://linear.app/api3-dao/issue/DAO-214/rename-updateimmutablycurried-function

Siegrift (approved):
👍 LGTM

PR #338, opened 2022-09-28
DAO-212 Add search and pagination to My Claims page

mcoetzee:
### What does this change?
- adds search to My Claims page (you can search for claim ID and policy metadata)
- adds pagination to My Claims page

The same SearchForm and Pagination components that we use on the policies pages (My Policies and Policy Select) were used for the My Claims page. These components were moved out of the `pages/components/policies` folder.

### Screenshots
Screenshot 2022-09-28 at 14 37 27
Screenshot 2022-09-28 at 14 37 59


src/components/input/input.module.scss
```diff
@@ -26,6 +27,7 @@
     }
   }
   input {
+    @include input-autofill;
```
Siegrift, line 30: 👍 Nice.

src/pages/my-claims/my-claims.tsx
```diff
@@ -14,12 +14,14 @@ import { isActive, useUserClaims } from '../../logic/claims';
 import { useUserPolicies, isActive as isPolicyActive } from '../../logic/policies';
 import styles from './my-claims.module.scss';
 
+type Filter = 'active' | 'inactive' | 'none' | null;
```
Siegrift, line 17: Does it make sense to have `none` filter?

Siegrift (approved):
👍 LGTM

src/pages/my-claims/my-claims.tsx
mcoetzee: Yeah, the `none` query param is for when both checkboxes are unchecked, and null (i.e. no query param) is for when both checkboxes are checked

Siegrift: yeah, but if there is none you will not see any results... But I guess it's fine since it's the intuitive behaviour.

PR #339, opened 2022-09-30
DAO-211 Remove ExternalLink component styling and reuse it

mcoetzee:
### What does this change?
The font-size and hover styles that came with the ExternalLink component made it a bit difficult to reuse across the different places. So this PR is about removing those styles to make the component easy to reuse. The small font-size and hover styles were added to the few usage sites that needed them.

Additionally, the ExternalLink props were extended to accept most link props.


mcoetzee:
> Have you checked if the old usage styling is unchanged?

Yip all usages have been tested and fixed where necessary 👍

PR #340, opened 2022-09-30
DAO-214 Rename updateImmutablyCurried to produceState and overload it

mcoetzee:
### What does this change?
- renames `updateImmutablyCurried` to `produceState`
- overloads the `produceState` function to make it consistent with Immer's `produce` function signature. Immer's `produce` function  creates a curried function when you only provide it a function, so it stands to reason that our `produceState` function should also handle both regular and curried usages

#### Regular usage
```ts
    const state = { a: 123, b: 'hi' };
    const newState = produceState(state, (draft) => {
      draft.a = 789;
    });
```

#### Curried usage
```ts
    interface State {
      a: number;
      b: string;
    }
    type CurriedFn = (state: State) => State;
    const curried: CurriedFn = produceState((draft) => {
      draft.a = 789;
    });

    const state = { a: 123, b: 'hi' };
    const newState = curried(state);
```


src/chain-data/helpers.ts
```diff
@@ -1,23 +1,28 @@
...
 import { ChainData, initialChainData } from './state';
 import { convertToEnsName } from '../logic/proposals/encoding/ens-name';
 import { go } from '@api3/promise-utils';
 
-export const updateImmutably = <T>(state: T, updateCb: (immutableState: T) => void) => {
+type ProducerFn<T> = (state: Draft<T>) => void;
+
+export function produceState<T>(producerFn: ProducerFn<T>): (state: T) => T;
+export function produceState<T>(state: T, producerFn: ProducerFn<T>): T;
+export function produceState<T extends object>(stateOrProducerFn: T | ProducerFn<T>, producerFn?: ProducerFn<T>) {
```
Siegrift, line 13: Btw. when I am doing the fn overloading, I just type the arguments as `any` (ususally the types just add more confusion).

PR #343, opened 2022-10-07
DAO-206 Add Execute Payout action for when the bot fails to execute the ruling for the user

mcoetzee:
The bot should take care of executing the ruling for the user, so this Execute Payout action is needed for the when the bot fails to do its job (a corner case).

### What does this change?
- adds a Execute Payout action when the arbitrator gave a `PayClaim` or a `PaySettlement` ruling, and the dispute is in the `Execution` period
- adds a `pass-dispute-period` Hardhat task in order to progress the dispute from the `Appeal` period to the `Execution` period

**Usage e.g.**
```
yarn claims:arbitrator-decide-to-pay-settlement --dispute-id 1 --appeal-period-length 0
yarn claims:pass-dispute-period --dispute-id 1
```

### DIFF
--- a/src/pages/claim-details/claim-actions.tsx
+++ b/src/pages/claim-details/claim-actions.tsx
@@ -101,7 +101,6 @@ export default function ClaimActions(props: Props) {
         transactions: [...transactions, { type: 'execute-claim-payout', tx }],
       });
       setStatus('submitted');
-      setModalToShow(null);
     } else {
       setStatus('failed');
     }
--- a/src/pages/claim-details/claim-actions.tsx
+++ b/src/pages/claim-details/claim-actions.tsx
@@ -93,6 +93,20 @@ export default function ClaimActions(props: Props) {
     }
   };
 
+  const handleExecutePayout = async () => {
+    setStatus('submitting');
+    const tx = await handleTransactionError(arbitratorProxy.executeRuling(claim.dispute!.id));
+    if (tx) {
+      setChainData('Execute claim payout transaction', {
+        transactions: [...transactions, { type: 'execute-claim-payout', tx }],
+      });
+      setStatus('submitted');
+      setModalToShow(null);
+    } else {
+      setStatus('failed');
+    }
+  };
+
   const handleModalClose = () => setModalToShow(null);
 
   switch (claim.status) {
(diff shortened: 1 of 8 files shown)


src/pages/claim-details/claim-details.tsx
```diff
@@ -158,7 +158,7 @@ function ClaimSummary(props: ClaimSummaryProps) {
       </div>
       <div className={styles.detailsItem}>
         <p className={globalStyles.bold}>Claim Created</p>
-        <p className={globalStyles.secondaryColor}>{format(claim.timestamp, 'do MMMM yyyy hh:mm')}</p>
+        <p className={globalStyles.secondaryColor}>{format(claim.timestamp, 'do MMMM yyyy HH:mm')}</p>
```
Siegrift, line 161: Can we maybe extract these to a constants and reuse? We are inlining these formats
image

src/pages/claim-details/claim-actions.tsx
```diff
@@ -261,22 +281,33 @@ export default function ClaimActions(props: Props) {
...
+                  </div>
+                  <p className={styles.actionMessage}>
+                    You can appeal within the given time frame or the counter offer will be automatically accepted.
+                  </p>
+                </>
+              )}
+              {dispute.period === 'Execution' && (
+                <div className={styles.actionPanel}>
+                  <Button variant="secondary" disabled={disableActions} onClick={handleExecutePayout}>
+                    Execute Payout
```
Siegrift, line 307: Shouldn't the user know that there will be a bot executing these rulings and this is only useful if the bot is down (or they want to execute the ruling immediately)?

Siegrift (approved):
👍 LGTM

src/pages/claim-details/claim-actions.tsx
mcoetzee: > or they want to execute the ruling immediately?

Correct, the user should be able to execute the ruling themselves if the bot is delayed, so I don't think we need to warn them about the bot.

src/pages/claim-details/claim-details.tsx
mcoetzee: One pain of doing this is that there are no natural names for these formats. E.g. here are some (probably bad) names that I came up with:
```ts
export const FULL_DATE_FORMAT = 'do MMMM yyyy';
export const COMPACT_DATE_FORMAT = 'dd MMM yyyy';
```
Now say we want to use `do MMM yyyy`, what would we call it?

But ja, I'm not too fussed, did you have specific naming in mind?

mcoetzee: Merging so long. Depending on the outcome of this discussion, we can create a new task for it.

PR #348, opened 2022-10-27
DAO-159 Update claim list items on My Claims page

mcoetzee:
### What does this change?
- updates the claim list items to align with the latest design, which includes:
  - using the policy metadata as claim title
  - new claim status wording
  - showing the claim ID
  - showing the claim created at date
  - showing pills for available actions that corresponds to the available actions on the claim's details page
- adds responsive layout for mobile
- extends `abbrStr` util function to handle custom start and end lengths

#### At full width
Screenshot 2022-10-27 at 17 30 44

#### When horizontal space gets tight for active claims' "Created" dates
Screenshot 2022-10-27 at 17 31 55

#### When horizontal space is too little to accommodate active claims' "Created" dates
Screenshot 2022-10-27 at 17 32 25

#### Mobile
Screenshot 2022-10-27 at 17 54 36


src/chain-data/helpers.ts
```diff
@@ -55,8 +55,10 @@ export const getNetworkData = async (provider: ethers.providers.Web3Provider | n
   return networkData;
 };
 
-export const abbrStr = (str: string) => {
-  return str.substr(0, 9) + '...' + str.substr(str.length - 4, str.length);
+export const abbrStr = (str: string, options?: { startLength?: number; endLength?: number }) => {
+  const startLength = options?.startLength ?? 9;
+  const endLength = options?.endLength ?? 4;
+  return str.substring(0, startLength) + '...' + str.substring(str.length - endLength, str.length);
```
mcoetzee, line 61: `substr` is deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

src/components/notifications/notifications.tsx
```diff
@@ -37,9 +37,22 @@ interface ToastPropsWithType extends ToastProps {
...
+        return images.success;
+      case 'warning':
+        return images.warning;
+    }
+  };
+
   return (
     <div className={classNames(styles.notificationBody, { [styles.url]: url })}>
-      <img src={`/${type}.svg`} alt={`${type} icon`} />
+      <img src={getIcon()} alt={`${type} icon`} />
```
Siegrift, line 55: Isn't `images[type]` enough?

Siegrift (approved):
Thanks for the PR description.
👍 LGTM

src/components/notifications/notifications.tsx
mcoetzee: It would be an improvement yes 👍, but similarly to `/${type}.svg`, it doesn't show up when searching for usages of the warning, error, info, or success images. Being able to search for usages will help if/when we convert those images to inline SVGs.

Siegrift: Makes sense, feel free to leave is as.

PR #350, opened 2022-11-02
DAO-227 Validate evidence IPFS hash with multiformats library

mcoetzee:
### What does this change?
It installs the [multiformats](https://www.npmjs.com/package/multiformats) library (245k weekly downloads at the time of PR) and uses it to validate the evidence IPFS hash (aka Content ID or CID).

I installed `multiformats` v9 (currently on v10) as v10 is not compatible with `react-scripts` v4. `react-scripts` v5 should be ok.
The library adds about 4.9kB gzipped JS to our bundle.


Siegrift (approved):
👍 LGTM

src/pages/new-claim/new-claim-form.tsx
```diff
@@ -19,15 +20,16 @@ export type FormStatus = 'idle' | 'validation_failed' | 'submitting' | 'submitte
 function getValidationMessages(form: FormState, policy: Policy) {
   const messages: ValidationMessages<FormState> = {};
 
-  if (form.evidence.trim().length === 0) {
-    messages.evidence = 'Please fill in this field';
+  const evidenceResult = goSync(() => CID.parse(form.evidence.trim()));
```
Siegrift, line 23: Judging from the fact that the validation is sync, it seems it doesn't check whether it actually points to an existing file.

mcoetzee: Correct. It's very unlikely that a user will accidentally enter a valid CID which doesn't exist.

PR #356, opened 2022-11-07
DAO-225 Remove My Policies page

mcoetzee:
### What does this change?
It removes the My Policies page. The API3 market will have the page that lists all the user's policies.


src/components/menu/menu.tsx
```diff
@@ -42,14 +41,11 @@ export const DesktopMenu = () => {
...
-          <div className={styles.menuActiveLine} />
-          <PoliciesIcon aria-hidden />
-          <p className={classNames(styles.menuItemText, globalStyles.textSmall)}>My Policies</p>
-        </div>
-      </NavLink>
-      <NavLink activeClassName={styles.menuActiveItem} to="/claims">
+      <NavLink
+        activeClassName={styles.menuActiveItem}
+        to="/claims"
+        isActive={() => pathname.startsWith('/claims') || pathname.startsWith('/policies')}
```
Siegrift, line 47: Why also `/policies`? For legacy reasons?

Siegrift (approved):
👍 LGTM

src/components/menu/menu.tsx
mcoetzee: So that it highlights the My Claims nav item as active when we are on the Policy Details and New Claim pages (`/policies/<ID>`,  `/policies/<ID>/claims/new`), now that we don't have the My Policies nav item anymore.

PR #357, opened 2022-11-09
DAO-224 Improve page transition scrolling for mobile

mcoetzee:
### What does this change?
- adds a scroll to top effect when moving between the New Claim steps
- adds a scroll to top effect for the Policy Select page
- uses the `useLayoutEffect()` hook to trigger the scroll (to avoid occasional flicker of "unscrolled" content)


Siegrift (approved):
👍 LGTM, also thx for fixing useEffect -> useLayoutEffect

PR #361, opened 2022-11-15
DAO-160 Cypress tests: Claim creation

mcoetzee:
### What does this change?
This PR only focuses on claim creation. It:
- adds a test that starts on the My Claims page, and goes through the whole New Claim flow (it includes input validation tests)
- adds a test for the Policy Select page (when the user has multiple active policies)
- adds a test for an inactive policy (you should not be able to create a claim)
- configures a Cypress `baseUrl`, to make `cy.visit()` a bit easier to use

More PRs will be created for testing the claim process etc. and for Percy snapshots


cypress/support/commands.ts
```diff
@@ -50,7 +50,7 @@ Cypress.Commands.add('login', () => {
     // Simulate injected metamask metamask provider
     (win as any).ethereum = ethersProvider;
   });
-  cy.visit(HOME_PAGE); // This is noop ife we are already on this page
+  cy.visit('/'); // This is noop ife we are already on this page
```
Siegrift, line 53: nit:

```suggestion
  cy.visit('/'); // This is noop if we are already on this page
```

src/pages/claim-details/claim-actions.tsx
```diff
@@ -119,7 +119,8 @@ export default function ClaimActions(props: Props) {
           return (
             <div className={styles.actionSection}>
               <p className={styles.mediator} data-testid="status-prefix">
-                <Api3Icon aria-hidden /> API3 Mediators
+                <Api3Icon aria-hidden />
```
Siegrift, line 122: Mhh, some prettier issue?

Siegrift (approved):
👍 Looks great.

src/pages/claim-details/claim-actions.tsx
mcoetzee: This was just to get rid of the extra white space (after the icon). It was a bit annoying to deal with in the Cypress tests 👍

PR #362, opened 2022-11-18
DAO-230 Cypress tests: Claim process (without arbitration)

mcoetzee:
### What does this change?
- adds a test for when the API3 Mediators accept the claim
- adds tests for when the API3 Mediators propose a settlement
- adds a test for when the API3 Mediators reject the claim (by not responding within 72 hours)

This PR does not include the dispute process/arbitration (that will come in a future PR).

I first added a new Hardhat task for creating a claim, but it tuned out to be significantly slower than visiting `/policies/${policyId}/claims/new` and using the frontend to submit the claim (around 8 seconds vs around 2 seconds while running Cypress locally).

### DIFF
--- a/cypress/e2e/claim-process.cy.ts
+++ b/cypress/e2e/claim-process.cy.ts
@@ -94,7 +94,7 @@ describe('Claim process', () => {
     });
 
     it('times out when the user does not act within 72 hours', () => {
-      cy.increaseTimeAndRelogin(259200); // 3 days
+      cy.increaseTimeAndRelogin(3 * 24 * 60 * 60); // 3 days
 
       // My Claims page
       cy.findByRole('link', { name: 'My Claims' }).filter(':visible').click();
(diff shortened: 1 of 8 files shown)


cypress/e2e/claim-process.cy.ts
```diff
@@ -0,0 +1,188 @@
...
+
+      // Go back to My Claims page
+      cy.findByRole('button', { name: 'Back' }).click();
+      cy.findByRole('heading', { name: 'My Claims' });
+      cy.findByTestId('claim-status').should('have.text', 'API3 Mediators (settled)');
+      assertClaimIsInactive();
+    });
+
+    it('times out when the user does not act within 72 hours', () => {
+      cy.increaseTimeAndRelogin(259200); // 3 days
```
Siegrift, line 97: Nit: 3 * 24 * 60 * 60

Siegrift (approved):
👍 LGTM

PR #364, opened 2022-11-24
DAO-232 Cypress tests: Policy updates (upgrades, downgrades, metadata updates)

mcoetzee:
### What does this change?
- adds a test for policy downgrades and upgrades
- adds a test for metadata updates


Siegrift (approved):
👍 LGTM

PR #366, opened 2022-11-28
DAO-226 Claim list: Alternative mechanism for showing the evaluation time for API3/Kleros

mcoetzee:
### What does this change?
- it calculates a deadline for Kleros evaluation
- it uses a different mechanism for showing the remaining evaluation time (on the claim list) to better communicate that the deadline is on API3/Kleros' side
- it removes the "created at" date from the claim list items

Kleros gives their ruling after the `Vote` period, so when we are in the `Evidence` period, we get the vote period end date by adding the vote period to the evidence period (the commit period is skipped because the sub court does not have hidden votes).

```ts
// Evidence period
deadline = addSeconds(dispute.periodChangedAt, evidencePeriodLength + votePeriodLength);
```
The period lengths are provided by the sub court (`subCourt.timesPerPeriod[]`)

### Screenshots
Screenshot 2022-11-28 at 15 05 16
Screenshot 2022-11-28 at 13 30 55

### DIFF
--- a/src/components/timer/timer.tsx
+++ b/src/components/timer/timer.tsx
@@ -26,37 +26,10 @@ interface Countdown {
 }
 
 const Timer = (props: Props) => {
-  const { deadline, size = 'normal', showDeadline, onDeadlineExceeded } = props;
-  const [countdown, setCountdown] = useState<Countdown>(() => calculateCountdown(deadline));
-
-  useEffect(() => {
-    const timer = setInterval(() => {
-      const countdown = calculateCountdown(deadline);
-      setCountdown(countdown);
-
-      if (countdown.dateDiff === 0) {
-        clearInterval(timer);
-      }
-    }, 1000);
-
-    return () => {
-      clearInterval(timer);
-    };
-  }, [deadline]);
-
-  const callbackRef = useRef(onDeadlineExceeded);
-  useEffect(() => {
-    callbackRef.current = onDeadlineExceeded;
-  });
+  const { deadline, size = 'normal', showDeadline } = props;
+  const countdown = useCountdown(deadline, props.onDeadlineExceeded);
 
   const { dateDiff } = countdown;
-  const previousDateDiff = usePrevious(dateDiff) || 0;
-  useEffect(() => {
-    // We trigger the callback when we have gone past the deadline (i.e. from a positive diff to a zero diff)
-    if (previousDateDiff > 0 && dateDiff === 0) {
-      callbackRef.current && callbackRef.current();
-    }
-  }, [previousDateDiff, dateDiff]);
 
   const largeSize = size === 'large' ? `${styles.large}` : '';
   const status = dateDiff > 0 ? `Remaining${showDeadline ? '' : ':'}` : 'Ended';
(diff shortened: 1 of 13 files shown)


src/components/timer/timer.tsx
```diff
@@ -100,6 +73,41 @@ const Timer = (props: Props) => {
 
 export default Timer;
 
+export function useCountdown(deadline: Date, onDeadlineExceeded?: () => void) {
```
Siegrift, line 76: Shouldn't we move this to hooks file?

src/logic/claims/data.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      dispute: null,
+    });
+
+    expect(deadline).toEqual(addDays(statusUpdatedAt, 3));
+  });
+
+  describe('when the dispute is in its "Evidence" period', () => {
+    it('returns the vote period end date', () => {
+      const periodChangedAt = addSeconds(new Date(), -42);
+      const evidencePeriod = 3.25 * 24 * 60 * 60;
```
Siegrift, line 30: Just curious, why the decimal point numbers?

src/logic/claims/data.ts
```diff
@@ -191,15 +191,26 @@ async function loadClaims(
...
     case 'ClaimCreated':
     case 'SettlementProposed':
       return addDays(claim.statusUpdatedAt, 3);
     case 'DisputeCreated':
-      return dispute?.period === 'Appeal' ? dispute.periodEndDate! : null;
+      switch (dispute?.period) {
+        case 'Evidence':
+          // Kleros gives their ruling after the voting period, so we get the voting period end date by adding the vote
+          // period to the evidence period (the commit period is skipped because the sub court does not have hidden votes).
+          return addSeconds(dispute.periodChangedAt, dispute.timesPerPeriod[0]! + dispute.timesPerPeriod[2]!);
```
Siegrift, line 206: I find these indexes a bit error-prone. Can we maybe have a fields such as `evidence`, 'commit', etc... instead of `timesPerPeriod` ? Alternatively, we could use named constants for indexing the times per period, but I like that a bit less.

src/logic/claims/data.test.ts
mcoetzee: That is the value on mainnet 👍

src/logic/claims/data.ts
mcoetzee: Yeah I was also on the fence about this. I reckon it will help to change:
```ts
      timesPerPeriod: subCourt.timesPerPeriod.map((time) => time.toNumber()),
```
to:
```ts
      periodTimes: {
        evidence: subCourt.timesPerPeriod[0].toNumber(),
        vote: subCourt.timesPerPeriod[2].toNumber(),
        appeal: subCourt.timesPerPeriod[3].toNumber(),
      },
```
when loading the data so that subsequent uses are at least a bit more obvious.

Siegrift: Yeah and if you want to stress out the order of the periods, you can write a comment on the interface type (but I don't think it's necessary).

src/logic/claims/data.test.ts
```diff
@@ -117,12 +114,11 @@ describe('calculateDeadline', () => {
           ruling: 'DoNotPay',
           period: 'Execution',
           periodChangedAt: new Date(),
-          timesPerPeriod: [
-            3.25 * 24 * 60 * 60, // Evidence
-            1000, // Commit
```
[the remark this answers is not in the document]
mcoetzee: The `Commit` period only applies to courts with hidden votes and our subcourt doesn't use hidden votes, so we can safely ignore it.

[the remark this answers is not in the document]
Siegrift: Mhh, but why was the test using it before?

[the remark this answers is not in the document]
mcoetzee: Oh because we basically saved the `timesPerPeriod` array as is (from the `getSubcourt()` SC call). This has now been changed.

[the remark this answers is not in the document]
Siegrift: Oh, I see. Thanks.

Siegrift (approved):
👍 LGTM

PR #360, opened 2022-11-15
[production] Add docker build and push CI workflow

dcroote:
Same as #359 (including image name & tag feedback updates) but for the `production` branch

PR #367, opened 2022-11-30
Add prebuilt Docker instructions to README

dcroote:
Updates the Action as in #365 and also updates the README to describe how to use the prebuilt Docker image.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -18,7 +18,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: ${{ secrets.DOCKERHUB_USERNAME }}/dao-dashboard
+          images: api3/dao-dashboard
           tags: |
             # set latest tag for production branch
             type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'production') }}
--- a/README.md
+++ b/README.md
@@ -11,8 +11,7 @@ need `docker` and, optionally, `git`.
 
 There are two approaches to running the dashboard locally. The first involves using the prebuilt Docker image, while the
 second involves building the image from source. In either case, the end result is a API3 dashboard running on port 7770
-of your localhost where it is safe to connect your wallet. Once you are finished interacting with the dashboard, the
-container can be stopped using `docker stop api3-dao-dashboard`.
+of your localhost where it is safe to connect your wallet.
 
 ### Running the prebuilt image
 
--- a/README.md
+++ b/README.md
@@ -4,20 +4,35 @@ The implementation of the DAO dashboard.
 
 [![ContinuousBuild](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml/badge.svg?branch=main)](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml)
 
-## Running dashboard on Mainnet in docker container
+## Running the dashboard on Mainnet in a docker container
 
-The decentralized approach of being a DAO member is to run API3 dashboard on your local machine. For that you will just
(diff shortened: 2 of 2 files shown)


mcoetzee (approved):
Nice one 👍

.github/workflows/build-docker.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: ${{ secrets.DOCKERHUB_USERNAME }}/dao-dashboard
+          images: api3/dao-dashboard
```
Siegrift, line 21: Shall this be also replaced on line 30? Why is it actually replaced?

README.md
```diff
@@ -4,21 +4,37 @@ The implementation of the DAO dashboard.
...
 
-The decentralized approach of being a DAO member is to run API3 dashboard on your local machine. For that you will just
-need `git` and `docker`. Note, that the production source code is on the `production` branch.
+The decentralized approach to being a DAO member is to run API3 dashboard on your local machine. For that you will just
+need `docker` and, optionally, `git`.
 
+There are two approaches to running the dashboard locally. The first involves using the prebuilt Docker image, while the
+second involves building the image from source. In either case, the end result is a API3 dashboard running on port 7770
+of your localhost where it is safe to connect your wallet. Once you are finished interacting with the dashboard, the
+container can be stopped using `docker stop api3-dao-dashboard`.
```
Siegrift, line 15: For me, it runs in the terminal window until you stop it via ctrl+C (or cmd+c). I don't see why is this needed.

README.md
```diff
@@ -4,21 +4,37 @@ The implementation of the DAO dashboard.
...
+
+### Running the prebuilt image
+
+The simplest, though less trustless, way to run the dashboard is using the prebuilt Docker image. This image is built
```
Siegrift, line 19: `though less trustless` sounds too fearful. I am not sure about this, since we are OK with using Airnode via Docker and here we have similar risks (Dockerhub being malicious which realistically won't happen).

Siegrift (commented):
We are discussing whether or not to move claims to a separate branch in https://api3workspace.slack.com/archives/C02ALA11APM/p1670061577550849

This is something we would like to have on the "default" branch (the branch you see when you go to the GH of a project) and also in production. So we might hold off merging this for a bit. 

cc: @mcoetzee

.github/workflows/build-docker.yml
dcroote: Aaron created a new user, `api3daodashboard`, specifically for this repo ([see here](https://github.com/api3dao/api3-dao-dashboard/pull/360#issuecomment-1325386654)) hence the image is `api3` to correspond with the repo and the username is passed as a secret (could be hardcoded as well, but no need).

README.md
dcroote: Confirmed- that instructive text was leftover from before and is outdated now- I removed it in the latest commit.

README.md
dcroote: I generally agree here. We could add `slightly` to lessen the frightfulness of the language or remove it.

dcroote:
> This is something we would like to have on the "default" branch (the branch you see when you go to the GH of a project) and also in production. So we might hold off merging this for a bit.

I'm not sure how the location of claims or the content of a default branch prevents updating the main README with instructions for using a docker image built from the production branch?

Siegrift:
> I'm not sure how the location of claims or the content of a default branch prevents updating the main README with instructions for using a docker image built from the production branch?

Because we might change what's the default branch and this should imo be in there. Was just an idea though... We might as well merge it to main.

README.md
dcroote: Removed

PR #373, opened 2023-01-03
Upgrade React to v18

mcoetzee:
### What does this change?
- upgrades React to v18
- bumps React related dependencies
- removes duplicate dashboard data load from the `Delegation` component (the data already gets loaded in the parent component) 

React 18 strict mode triggers `useEffect` and `useMemo` hooks twice when mounting (in development). This results in a pretty poor developer experience as it causes our data hooks to trigger 3 times on mount (the third trigger is caused by the `useMemo`'ed smart contracts which are included in the `useEffect` dependency arrays). For me, it feels like the DX hit is too large, so this PR turns off strict mode.


src/index.tsx
```diff
@@ -30,9 +29,6 @@ if (process.env.REACT_APP_NODE_ENV === 'development' && (window as any).ethereum
...
 
-ReactDOM.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-  document.getElementById('root')
-);
+const root = createRoot(document.getElementById('root')!);
+
+root.render(<App />);
```
Siegrift, line 34: I support your decision about not using strict mode, but add a comment about that. We will forget about this after some time and wonder.

Siegrift (approved):
👍 LGTM

src/index.tsx
mcoetzee: Good idea 👍

PR #389, opened 2023-01-10
Generate typechain artifacts with ABIs

mcoetzee:
### What does this change?
It replaces the hacky approach to generating typechain artifacts because it is causing the build to fail in our lockfile maintenance PR: https://github.com/api3dao/api3-dao-dashboard/pull/388 

It replaces the approach by adding the contracts' ABI files and using them to generate the typechain artifacts.

Once this PR is merged, then the lockfile maintenance build will be able to run.


Siegrift (approved):
👍 Good idea to use only the ABIs only. LGTM.

PR #404, opened 2023-03-06
Update readme

Siegrift:
Based on the discussion in https://api3workspace.slack.com/archives/C02ALA11APM/p1678093121848019

The README recommends to update the `.eth.link` domain, which is no longer owned by Cloudflare and [should not be used anymore](https://twitter.com/luigyGT/status/1554092735229034496). MetaMask seems to [block](https://github.com/MetaMask/eth-phishing-detect/issues/11814) these URLs already. I've removed this part of docs.

While reading the README I got a feeling we are mixing content for two audiences, tech savvy users and developers. I've decided to move developer instructions to separate dev-README. I still think keeping the `main` (not production) branch as a default branch is good idea. Wdyt?

I've also updated the GH "about section" to use [api3.eth/](http://api3.eth/). Btw. for this I had to edit the input type to text, because it is not a valid URL according to type URL 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,13 @@ The implementation of the DAO dashboard.
 
 [![ContinuousBuild](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml/badge.svg?branch=main)](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml)
 
+We use the `main` branch to develop new features. For production code, see the
+[the production branch](https://github.com/api3dao/api3-dao-dashboard/tree/production).
+
+## Development instructions
+
+You can find development instructions in a separate [dev-README](./dev-README.md).
+
 ## Running the dashboard on Mainnet in a Docker container
 
 The decentralized approach to being a DAO member is to run API3 dashboard on your local machine. For that you will just
@@ -12,6 +19,8 @@ need `docker` and, optionally, `git`.
 There are two approaches to running the dashboard locally. The first involves using the prebuilt Docker image, while the
 second involves building the image from source. In either case, the end result is a API3 dashboard running on port 7770
 of your localhost where it is safe to connect your wallet.
+
+Currently, it's only possible to build the Docker image on a UNIX-like OS. If you use Windows, you can use WSL2.
 
 ### Running the prebuilt image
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,8 +1,13 @@
-# api3-dao-dashboard
+# api3-dao-dashboard [![ContinuousBuild](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml/badge.svg?branch=main)](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml)
```
dcroote, line 1: I'd prefer the badge under the title rather than in it.

README.md
```diff
@@ -1,8 +1,13 @@
...
 
 The implementation of the DAO dashboard.
 
-[![ContinuousBuild](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml/badge.svg?branch=main)](https://github.com/api3dao/api3-dao-dashboard/actions/workflows/main.yml)
+We use the `main` branch to develop new features. For production code, see the
+[production branch](https://github.com/api3dao/api3-dao-dashboard/tree/production).
+
+## Development instructions
+
+You can find development instructions in a separate [dev-README](./dev-README.md).
```
dcroote, line 10: I think this is better further towards the bottom so that users see instructions for running the container first.

dcroote (approved):
Looks good, 2 nits from me.

> While reading the README I got a feeling we are mixing content for two audiences, tech savvy users and developers. I've decided to move developer instructions to separate dev-README.

Makes sense.

> I still think keeping the main (not production) branch as a default branch is good idea. Wdyt?

Agreed

> I've also updated the GH "about section" to use [api3.eth/](http://api3.eth/).

👍

README.md
Siegrift: My intention was to save vertical space. I can revert though if you still prefer it separate.

README.md
mcoetzee: I think I prefer how Emmanuel has it now, i.e. having it close to the top.

mcoetzee (approved):
> I still think keeping the main (not production) branch as a default branch is good idea. Wdyt?

Agree 👍

README.md
dcroote: I would argue we should cater to users first then devs, but if you both like it here then ok.

README.md
dcroote: I haven't seen many repos with it like this, but ok, I don't feel that strongly.

PR #411, opened 2023-05-09
Allow simple ETH transfer proposals

Siegrift:
See: https://api3workspace.slack.com/archives/C02ALA11APM/p1683621470320789

## Rationale

Review commit by commit.

Initially, we used only a very loose validation of the proposal form. Later we improved this, but we also prevented a use case of simple ETH transfer. 

To support this use case, I've added special cases to the validation. If both target contract signature and parameters are empty, the validation passes and a proposal can be created. Since signature and parameters can be empty now, I think it makes sense to hide them in proposal details.

I've tested it manually locally, but creating e2e is complex so I opted for encoding unit tests only.

## Images

1. I filled the proposal form
image

2. After creating it, it got immediately executed (since I have all the voting power)
image

3. Proposal details
image

### DIFF
--- a/src/logic/proposals/encoding/encoding.test.ts
+++ b/src/logic/proposals/encoding/encoding.test.ts
@@ -119,6 +119,20 @@ describe('encoding incorrect params', () => {
       new EncodedEvmScriptError('targetValue', 'Value must be greater than 0 for ETH transfers')
     );
   });
+
+  it('empty array parameters ("[]") for simple ETH transfer', async () => {
+    const invalidData = updateImmutably(newFormData, (data) => {
+      data.parameters = '[]';
+      data.targetSignature = '';
+    });
+
+    const goRes = await goEncodeEvmScript(mockedProvider, invalidData, api3Agent);
+
+    assertGoError(goRes);
+    expect(goRes.error).toEqual(
+      new EncodedEvmScriptError('parameters', 'Please specify an empty string for parameters for simple ETH transfer')
+    );
+  });
 });
 
 describe('encoding invalid target signature', () => {
@@ -157,7 +171,7 @@ describe('encoding invalid target signature', () => {
 
     assertGoError(goRes);
     expect(goRes.error).toEqual(
-      new EncodedEvmScriptError('parameters', 'Please specify the correct number of function arguments')
+      new EncodedEvmScriptError('parameters', 'Please specify an empty string for parameters for simple ETH transfer')
     );
   });
 });
--- a/src/logic/proposals/encoding/encoding.test.ts
+++ b/src/logic/proposals/encoding/encoding.test.ts
@@ -35,14 +35,14 @@ const mockedProvider: providers.JsonRpcProvider = {
   },
 } as any;
 
-test('correct encoding', async () => {
+test('correct contract call proposal', async () => {
   const goRes = await goEncodeEvmScript(mockedProvider, newFormData, api3Agent);
 
   assertGoSuccess(goRes);
   expect(goRes.data).toBeDefined();
 });
 
-it('allows simple ETH transfers using zero length parameters and empty target signature', async () => {
+test('simple ETH transfers using zero length parameters and empty target signature', async () => {
   const validData = updateImmutably(newFormData, (data) => {
     data.parameters = '[]';
     data.targetSignature = '';
@@ -54,8 +54,8 @@ it('allows simple ETH transfers using zero length parameters and empty target si
   expect(goRes.data).toBeDefined();
 });
 
-describe('encoding incorrect params', () => {
-  it('incorrect parameter values', async () => {
+describe('encoding incorrect parameters', () => {
+  test('incorrect parameter values', async () => {
     const invalidData = updateImmutably(newFormData, (data) => {
       data.parameters = JSON.stringify([123, 'arg1']); // they are in the wrong order
     });
@@ -68,7 +68,7 @@ describe('encoding incorrect params', () => {
     );
   });
 
-  it('wrong shape', async () => {
+  test('wrong shape', async () => {
     const invalidData = updateImmutably(newFormData, (data) => {
       data.parameters = JSON.stringify({ param: 'value' });
     });
@@ -79,7 +79,7 @@ describe('encoding incorrect params', () => {
     expect(goRes.error).toEqual(new EncodedEvmScriptError('parameters', 'Make sure parameters is a valid JSON array'));
   });
 
-  it('wrong number of parameters', async () => {
+  test('wrong number of parameters', async () => {
     const invalidData = updateImmutably(newFormData, (data) => {
       data.parameters = JSON.stringify(['arg1']);
     });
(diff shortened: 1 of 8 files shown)


Siegrift (commented):
Few notes:
1. I needed to use node v14 to deploy the DAO contracts, otherwise I was getting some strange errors. There might be a time when the app becomes incompatible with Node v14 and we will have to make the changes in DAO contracts.
2. It feels a bit wierd that only "parameters" have a default value in the new proposal form.

bbenligiray:
I'd expect that parameters would need to be `[]` instead of null

Siegrift:
> I'd expect that parameters would need to be [] instead of null

The thing is that `[]` has already a well defined meaning - empty parameters for a contract call. I personally like this distinction.

mcoetzee (changes requested):
I've mainly got a question around whether it makes sense to allow a zero value for these simple ETH transfer proposals. I was able to create one locally, and this is how the details page looks like (we hide the value when it is zero):
Screenshot 2023-05-10 at 11 13 56

Siegrift:
> I've mainly got a question around whether it makes sense to allow a zero value for these simple ETH transfer proposals. 

Good point. We probably should not.

bbenligiray:
`ethers.utils.defaultAbiCoder.encode([], [])` is `0x` and `ethers.utils.defaultAbiCoder.encode([], null)` throws. `null` is not a valid value for something to be ABI-encoded. Maybe have the whole parameters textbox disappear when the signature is empty.

> I've mainly got a question around whether it makes sense to allow a zero value for these simple ETH transfer proposals. 

There is an edge case where someone may want to send a value with some parameters encoded but no function specified so that the target's fallback function gets invoked. However, the current implementation doesn't allow that (because it wouldn't know how to encode the parameter values with an empty signature) so it would be a good idea to not allow 0 value when the signature is empty.

Siegrift:
> ethers.utils.defaultAbiCoder.encode([], []) is 0x and ethers.utils.defaultAbiCoder.encode([], null) throws. null is not a valid value for something to be ABI-encoded. 

I mean this is handled in the implementation. We never pass `null` during encoding.

> Maybe have the whole parameters textbox disappear when the signature is empty.

I think both `target contract signature` and `parameters` should be together so I dislike the suggestion. To most clear UI/UX to me is hide it behind [accordeon](https://mui.com/material-ui/react-accordion/) called "contract call parameters" which the user could expand if necessary. Not sure if it's worth implementing.

> There is an edge case where someone may want to send a value with some parameters encoded but no function specified so that the target's fallback function gets invoked.

Good point. Btw. this will not work with our metadata format even if we would like to support it and we would need a new version for this.

mcoetzee:
> To most clear UI/UX to me is hide it behind [accordeon](https://mui.com/material-ui/react-accordion/) called "contract call parameters" which the user could expand if necessary. Not sure if it's worth implementing.

Hmm seeing as these simple ETH transfer proposals isn't the norm, I would rather keep things as is.

bbenligiray:
> `ethers.utils.defaultAbiCoder.encode([], [])` is `0x` and `ethers.utils.defaultAbiCoder.encode([], null)` throws. `null` is not a valid value for something to be ABI-encoded.

I meant this in the sense that there is a convention for using `[]` when there is nothing to ABI-encode. Similarly, if you compile `contract EmptyContract {}` the contract ABI will be `[]`. Obviously these are not exactly the same, but it's enough for me to try putting in `[]` instead of leaving it empty on my first try (as you can see on my screenshot on Slack).

Siegrift:
> I meant this in the sense that there is a convention for using [] when there is nothing to ABI-encode. Similarly, if you compile contract EmptyContract {} the contract ABI will be []. Obviously these are not exactly the same, but it's enough for me to try putting in [] instead of leaving it empty on my first try (as you can see on my screenshot on Slack).

I understand your reasoning, but that comes form your deep understanding of how things work under the hood. When you create an ETH transfer proposal (and there is no contract function to be called) both `parameters` and `address` don't apply. I think with that perspective it's more reasonable to let them empty (not touch/fill them at all). But as I said, I can change that if you want it your way.

With this new implementation the validation would let you know that you want them to be empty. Does this sound OK ? 
image

bbenligiray:
> With this new implementation the validation would let you know that you want them to be empty.

It doesn't need to. It should error when parameters is empty and it should encode them as usual when it's `[]` (and get `0x` considering that `types` is also `[]`).

Siegrift:
> It doesn't need to. It should error when parameters is empty and it should encode them as usual when it's [] (and get 0x considering that types is also []).

Ok. I've changed the implementation to accept `[]` instead of empty string (see the image below).

image

src/logic/proposals/encoding/encoding.ts
```diff
@@ -112,6 +117,7 @@ export const goEncodeEvmScript = async (
       range(parameterTypes.length).map(async (i) => {
         const param = targetParameters[i]!;
         if (parameterTypes[i] !== 'address') return param;
+        if (typeof param !== 'string') throw new Error('Parameter must be an ENS string');
```
mcoetzee, line 120: The error message `Parameter must be an ENS string` makes it sound like we only accept ENS names here.

src/logic/proposals/encoding/encoding.test.ts
```diff
@@ -67,7 +79,33 @@ describe('encoding incorrect params', () => {
     expect(goRes.error).toEqual(new EncodedEvmScriptError('parameters', 'Make sure parameters is a valid JSON array'));
   });
 
-  it('empty parameters', async () => {
+  it('wrong number of parameters', async () => {
```
mcoetzee, line 82: I think these tests would read a bit better if they were named along the lines of:
```ts
it('returns an error when the number of parameters do not match', ...
it('returns an error for zero length parameters and a non-empty target signature', ...
...
```

mcoetzee (approved):
Nice one 👍. I've just got some minor feedback, so I'm approving so long.

bbenligiray:
Thanks

src/logic/proposals/encoding/encoding.ts
Siegrift: Mhh, but it will be only present if there is a non-numeric value passed as address. But I guess it doesn't hurt to reword this.

Siegrift: Actually, I remember why I didn't put much effort into this error message. This was mainly for type inference to pass string to `convertToAddressOrThrow`. Even if this throws we still use the generic `Ensure parameters match target contract signature` message. I think this is good enough, but I've added a test for this. If you disagree open an issue (it don't wanna implement it in this PR).

src/logic/proposals/encoding/encoding.test.ts
Siegrift: It reads better when you open this in IDE since there is more context provided by the wrapping `describe`. I've tried to improve the readability using `it` and `test` so that it reads better. But, personally I am not too keen on this.

Siegrift:
I should have squash merged - sorry.

PR #413, opened 2023-05-22
Upgrade web3modal to v2

### LINKED ISSUE #401: Upgrade Wallet Connect (web 3 modal) to version 2

mcoetzee:
Resolves #401 

### What does this change?
- upgrades the web3modal to v2, which introduces a whole new approach through the use of the `wagmi` package that provides various React hooks like `useAccount()`, `useSigner()`, etc. Requests that are made with the `provider` will now appear in the network tab
- removes the ENS name lookups from the proposal load (they were responsible for creating over 300 requests on the History page). Instead we do some preloading of the proposal creator, and defer the ENS name lookups for the target address and parameters (loaded on the Proposal Details page)
- removes the switch account code that I believe was there only for Hardhat. It requires more effort to implement which doesn't seem worth it at this stage
- makes use of more JS bundles (the use of `web3modal/react` package results in an async chunk/bundle)
- adds test utils that creates a mock wagmi client 

**Outstanding question:**
Do we still need to support Rinkeby? `wagmi` deprecated Rinkeby a while ago and doesn't export the chain anymore.
**Edit:** **Answer**: No, we don't support Rinkeby anymore. I've logged #414 to remove Rinkeby

### Screenshots
Screenshot 2023-05-22 at 10 05 37


config-overrides.js
```diff
@@ -1,45 +1,73 @@
...
+        );
+      },
+    ];
 
-  return config;
+    return config;
+  },
+
+  jest(config) {
+    config.transformIgnorePatterns = [
```
Siegrift, line 66: nit: I'd add an assert that verifies `transformIgnorePatterns` is what it is now. Since you are overriding the values, it might be pretty hard to debug in the future in case they modify its content.

src/chain-data/context.tsx
```diff
@@ -35,4 +37,37 @@ const ChainDataContextProvider = (props: { children: ReactNode }) => {
 
 export default ChainDataContextProvider;
 
-export const useChainData = () => useContext(ChainDataContext);
+export const useChainData = () => {
+  const data = useContext(ChainDataContext);
+  const { chain } = useNetwork();
+  const provider = useProvider({ chainId: chain?.id });
```
Siegrift, line 43: I believe you don't need to use `chainId`

src/chain-data/context.tsx
```diff
@@ -35,4 +37,37 @@ const ChainDataContextProvider = (props: { children: ReactNode }) => {
...
+  const { isConnected, address } = useAccount();
+
+  // Note: The signer is briefly undefined after connecting or after switching networks.
+  if (isConnected && address && signer) {
+    const networkName = updateNetworkName(chain?.network || '');
+    return {
+      ...data,
+      isConnected: true,
+      provider,
+      signer,
```
Siegrift, line 55: Since we are using wagmi, it doesn't make sense to have data like `signer`, `provider`, `userAccount` in our context. I expect it is a larger (and tedious) refactor to remove it - so better done in separate issue.

Siegrift: Nevermind. I see that you did something even better - modified the `useChainData` hook to return combination of our context and wagmi context. Very nice 🥂

src/components/sign-in/sign-in.tsx
```diff
@@ -1,52 +1,29 @@
...
-  const onDisconnect = () => {
-    if (provider) {
-      const externalProvider: any = provider.provider;
-      if (typeof externalProvider.close === 'function') {
-        externalProvider.close();
-      }
-    }
-    setChainData('User disconnected', initialChainData);
-  };
+  const { data: userAccountName } = useEnsName({ address: userAccount as Address });
```
Siegrift, line 26: Personal remark: I found it quite annoying how they strongly typed the `Address`.

src/contracts/network.ts
```diff
@@ -19,6 +19,7 @@ export const updateNetworkName = (networkName: string) => {
 export const getDaoAddresses = (networkName: string) => {
   switch (networkName) {
     case 'localhost':
```
Siegrift, line 21: Do we need `localhost` here?

src/logic/proposals/encoding/encoding.ts
```diff
@@ -224,7 +224,7 @@ export const decodeEvmScript = async (
       ['address', 'uint256', 'bytes'],
       utils.hexDataSlice(callData, 4)
     );
-    const targetContractAddress = await tryConvertToEnsName(provider, executionParameters[0]);
+    const targetContractAddress = executionParameters[0];
```
Siegrift, line 227: I believe this is wrong. The EVM encoding will fail if you try to encode an ENS name (e.g. "bbenligiray.eth") as `address`. So I think we need to convert to ENS here.

Siegrift: Also, this encoding is one of the most critical parts of the FE so I find it a bit strange/sad that no tests failed because of this change. Or am I missing something?

Siegrift (changes requested):
Did a first pass. Looks very good, but had some comments/questions.

config-overrides.js
```diff
@@ -1,45 +1,73 @@
...
+module.exports = {
+  webpack(config, env) {
+    if (env === 'production') {
+      // Webpack produces a single JS bundle with the default configuration (for the initial non-async chunk),
+      // so we specify "chunks: all" and provide a cache group to generate more bundles
+      config.optimization.splitChunks = {
+        chunks: 'all',
+        cacheGroups: {
+          // The "(?<!node_modules.*)" regex ignores nested copies of packages, so that they're bundled with their issuer.
+          // See: https://github.com/vercel/next.js/pull/9012
```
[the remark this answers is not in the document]
mcoetzee: Yeah I used the same cache groups for the market and it has been running in prod for a while: https://market.api3.org/

And you can see it in action for the dashboard: https://bafybeigmubosdbz5oask2a5induhj5vswclculxg7ysltlzeeuzwvnwr2u.on.fleek.co/#/board:

src/components/sign-in/sign-in.tsx
mcoetzee: Yeah I also found that a bit overkill.

src/contracts/network.ts
mcoetzee: Good point, I will remove 👍

src/logic/proposals/encoding/encoding.ts
mcoetzee: Hmm this is the `decodeEvmScript` function where the `tryConvertToEnsName` is used just for display purposes afaict. Are you thinking about the encoding function? That has not been touched. 

How I understand it, the encoding process makes sure any potential ENS names are converted to addresses (before creating the proposal), and when decoding, the addresses don't need to be converted into ENS names upfront, because we only convert them to ENS names for display purposes.

src/chain-data/context.tsx
mcoetzee: Removing them seems to work as expected 👍

Siegrift (approved):
👍 LGTM

PR #416, opened 2023-05-29
Add tests around encoding ENS names back in

mcoetzee:
#401 

### What does this change?
It adds the tests back that was removed as part of the web3modal upgrade here: https://github.com/api3dao/api3-dao-dashboard/commit/d203177610586429f9b7fb3978156d295230eb6f#diff-d007a916315fd637a98142838e0bff7a8336ac91e5d2210aa5e50cb0d0915408L248

I have realised that we don't need to remove them completely, but that they should be adjusted.


Siegrift (approved):
👍 LGTM

PR #418, opened 2023-06-20
Use v2 wallet connector (Wallet Connect)

mcoetzee:
### What does this change?
It bumps the Wallet Connect version to v2, but it wasn't as simple as updating the version from 1 to 2:
```diff
// src/wallet-connect.ts
-connectors: w3mConnectors({ version: 1, chains, projectId }),
+connectors: w3mConnectors({ version: 2, chains, projectId }),
```
It turns out, when using the v2 connector and you construct a smart contract with a `signer`, it fails when calling a method on the smart contract that loads data from the chain. **Note** this is when using a wallet (like Trust/Rainbow) via Wallet Connect, MetaMask works fine. To fix the data loading issue, it boils down to using the `provider` instead:
```diff
-ConvenienceFactory.connect(contracts.convenience, signer);
+ConvenienceFactory.connect(contracts.convenience, provider);
```
And connecting as the `signer` when creating transactions:
```diff
-api3Pool.stake(parsedValue))
+api3Pool.connect(signer).stake(parsedValue))
```

I managed to reproduce the error and I opened a [discussion](https://github.com/orgs/WalletConnect/discussions/2694) to get feedback whether this is expected behaviour or not (I suspect it's expected behaviour).

#### Side Note
The Market had a [weird bug](https://github.com/orgs/WalletConnect/discussions/2673) with the v2 connector that forces us to upgrade the `wagmi` dependency to v1 (that uses `viem`). The DAO Dashboard doesn't have that bug, so I decided to keep the `wagmi` dependency as is (at least for now).

### DIFF
--- a/src/contracts/hooks.ts
+++ b/src/contracts/hooks.ts
@@ -11,34 +11,54 @@ import {
 } from './artifacts/factories';
 import { initialChainData } from '../chain-data/state';
 
+const useContractReader = () => {
+  const { contracts, provider, signer } = useChainData();
+  const { connector } = useAccount();
+
+  /*
+   * Please note the following:
+   * 1. When connected via non-MetaMask wallets (like via Wallet Connect), loading contract data fails when
+   *    the smart contract was constructed with a signer.
+   * 2. Data loads a considerable amount faster with MetaMask when using the signer.
+   *
+   * In conclusion, we want to use the provider when connecting the contracts, except when we are connected via
+   * MetaMask (and aren't running the app locally).
+   */
+  const reader = connector?.name === 'MetaMask' && process.env.NODE_ENV !== 'development' ? signer : provider;
+  return {
+    contracts,
+    reader,
+  };
+};
+
 export const useApi3Pool = () => {
-  const { contracts, signer } = useChainData();
+  const { contracts, reader } = useContractReader();
 
   return useMemo(() => {
-    if (!contracts || !signer) return null;
-    return Api3PoolFactory.connect(contracts.api3Pool, signer);
-  }, [signer, contracts]);
+    if (!contracts || !reader) return null;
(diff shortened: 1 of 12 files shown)


bbenligiray (commented):
FWIW while using ethers, if a signer is created without the optional provider data, it can't be used to read from contracts. However, you can explicitly [connect](https://docs.ethers.org/v5/api/signer/#Signer-connect) it to a provider after creation, after which it can be used to read from contracts. So maybe the signer or the [contracts](https://docs.ethers.org/v5/api/contract/contract/#Contract-connect) here could have been connected to the provider to avoid the signer/reader distinction.

mcoetzee:
> FWIW while using ethers, if a signer is created without the optional provider data, it can't be used to read from contracts. However, you can explicitly [connect](https://docs.ethers.org/v5/api/signer/#Signer-connect) it to a provider after creation, after which it can be used to read from contracts. So maybe the signer or the [contracts](https://docs.ethers.org/v5/api/contract/contract/#Contract-connect) here could have been connected to the provider to avoid the signer/reader distinction.

Ah that is interesting. Will look into it now 👍

mcoetzee:
FYI @bbenligiray I didn't have success calling the `signer`'s `connect` function. It gives the error (it's a [JsonRpcSigner](https://docs.ethers.org/v5/api/providers/jsonrpc-provider/#JsonRpcSigner)):
`Error: cannot alter JSON-RPC Signer connection (operation="connect", code=UNSUPPORTED_OPERATION)`

I also tried the approach of using `provider.getSigner(address)` when connecting the contracts, but that also didn't work (it fails when creating a transaction).

So it seems that the approach in this PR is the best I got.

andreogle (approved):
Really impressive work on this 🙌🏻 . I don't have any better suggestions

bbenligiray:
No further comments from me

PR #424, opened 2023-10-02
Upgrade wagmi to v1

mcoetzee:
Fixes #425 

According to [this issue](https://github.com/WalletConnect/web3modal/issues/1345), the problem should be resolved by updating to the latest `wagmi`.

### What does this change?
- upgrades `wagmi` to v1, which comes with `viem`
- upgrades Typescript to v5 (`viem` depends on Typescript v5)
- shows a warning message when `localStorage.setItem()` fails

### How did you action this task?
- adapters provided by `wagmi` (https://wagmi.sh/react/ethers-adapters) were used to convert the `viem` public and wallet clients to the `ethers` provider and signer.
- I don't have full confidence that the original issue is resolved, so I thought it would be a good idea to let the user know if/when it happens again and to give them the ability to quickly sort it out.

### Screenshots
Screenshot 2023-10-03 at 15 43 59

PR #428, opened 2023-10-11
Upgrade wagmi to v1 (part 2)

mcoetzee:
Follow up to https://github.com/api3dao/api3-dao-dashboard/pull/424

### What does this change?
- updates the browserlist to filter out ancient browsers (IE and Android browser version <= 4.4.4) so that transpilation doesn't interfere with BigInts. The issue is described [here](https://github.com/hirosystems/stacks.js/issues/1096#issuecomment-946350299)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,8 @@
   "browserslist": {
     "production": [
       ">0.2%",
+      "not ie <= 99",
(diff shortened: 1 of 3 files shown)


andreogle (approved):
👍🏻

PR #431, opened 2023-11-04
Update wallet connect and use custom RPC

Siegrift:
Based on [user reported issue](https://forum.api3.org/t/dao-dashboard-update-oct-16-2023/2043/4) and [internal communication](https://api3workspace.slack.com/archives/C02ALA11APM/p1698983648604699).

## Rationale

The failure was caused by CloudFlare problems. Because the public RPC URL is CLoudFlare ETH it was down as well. 

One of the first issues I encountered was that regular MM connection was failing on localhost. The reason was a check which used public RPC provider instead of the provider bundled with MM on localhost. It's better to remove this special (localhost) handling.

I tried using [wagmi multiple providers configuration](https://wagmi.sh/react/providers/configuring-chains#multiple-rpc-providers) with or without [ranking](https://wagmi.sh/react/providers/configuring-chains#rank-optional), but both options produced error notifications. The site usually loaded, but I don't trust their [fallback transport implementation](https://viem.sh/docs/clients/transports/fallback.html). Ultimately, I suggest using only our alchemy provider URL.

I tested the implementation with Alchemy and whitelisted only API3 contracts and the site worked fine, except MM mobile which does some multicalls and I needed to whitelist [this contract](https://etherscan.io/address/0xca11bde05977b3631167028862be2a173976ca11) as well. I tested Wallet Connect on Trust Wallet and MetaMask.

Since I spent a lot of time fixing these issues I took the time and upgraded to wallet connect v3. The migration was quite simple because there is a [guide to follow](https://docs.walletconnect.com/web3modal/upgrade), but I needed to come up with metadata (used for display purposes in the mobile wallets when using wallet connect).


.env.example
```diff
@@ -11,3 +11,10 @@ REACT_APP_SENTRY_DSN=
 
 # Wallet Connect project ID
 REACT_APP_PROJECT_ID=e7e6e90622ec8e65c4d02e829a68529a
+
+# The RPC URL will be visible in the browser network tab (leaking the API key). It is advised to whitelist only API3 DAO
```
Siegrift, line 15: Specifically [these addreses](https://github.com/api3dao/api3-dao-dashboard/blob/551c112e553fd9243274a13247d60a62475f7178/src/contract-deployments/mainnet-dao.json#L5)

src/wallet-connect.ts
```diff
@@ -1,36 +1,69 @@
...
-        http: chain.rpcUrls.default.http[0]!,
-        webSocket: chain.rpcUrls.default.webSocket?.[0],
-      };
-    },
-  }),
-]);
-
-export const wagmiClient = createConfig({
+const chainInfos = [
+  { chain: mainnet, rpcUrl: process.env.REACT_APP_MAINNET_RPC_URL },
```
Siegrift, line 22: The advantage is that the RPC url is managed completely by us. There is still potential for this RPC URL to be down (limit exceeded or the provider having issues), but we can react more easily to this, e.g. by re-deploying the app with a different RPC.

src/wallet-connect.ts
```diff
@@ -1,36 +1,69 @@
...
+  { chain: hardhat, rpcUrl: 'http://localhost:8545' },
+];
+
+// The metadata is used by WalletConnect to display the app information in the user Wallet.
+const metadata = {
+  name: 'API3 DAO',
+  description: 'Decentralized APIs for Web 3.0', // Taken from API3 whitepaper.
```
Siegrift, line 29: All of the metadata fields are required, but nor MM nor Trust wallet seems to use this description.

package.json
```diff
@@ -56,8 +56,7 @@
   },
   "dependencies": {
     "@sentry/react": "^6.19.7",
-    "@web3modal/ethereum": "^2.7.1",
-    "@web3modal/react": "^2.7.1",
+    "@web3modal/wagmi": "^3.2.0",
```
andreogle, line 59: Nice one 👍🏻

.env.example
andreogle: Can we make this a bit more explicit in the comment? Maybe a link to the contract addresses or just put them here since they won't change? Right now it's not obvious where to find them

andreogle (approved):

.env.example
bbenligiray: Did this 
Screenshot 2023-11-06 at 10-39-35 Alchemy Dashboard

mcoetzee (commented):
Nice one 👍

FYI I think there is some more whitelisting to do, because I just ran this locally on mainnet and got some 403s from the provider:
Screenshot 2023-11-06 at 11 06 01

bbenligiray:
https://etherscan.io/address/0xc0497E381f536Be9ce14B0dD3817cBcAe57d2F62 seems to be the UniversalResolver contract for ENS, I whitelisted that as well. Oddly enough I can't find any reference to this address in ethers so I suspect that it may change if we bump the ethers version at some point.

.env.example
```diff
@@ -11,3 +11,10 @@ REACT_APP_SENTRY_DSN=
...
+# contracts making the RPC URL useless for others. To support Wallet Connect with MM wallet without errors, the
+# following multicall contract needs to be whitelisted as well:
+#
+# https://etherscan.io/address/0xca11bde05977b3631167028862be2a173976ca11
+REACT_APP_MAINNET_RPC_URL=
```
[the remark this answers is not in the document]
bbenligiray: It's already set, we had been using that for WalletConnect and it was never unset once we stopped doing so

.env.example
```diff
@@ -11,3 +11,10 @@ REACT_APP_SENTRY_DSN=
 
 # Wallet Connect project ID
 REACT_APP_PROJECT_ID=e7e6e90622ec8e65c4d02e829a68529a
+
+# The RPC URL will be visible in the browser network tab (leaking the API key). It is advised to whitelist only API3 DAO
+# contracts making the RPC URL useless for others. To support Wallet Connect with MM wallet without errors, the
+# following multicall contract needs to be whitelisted as well:
+#
+# https://etherscan.io/address/0xca11bde05977b3631167028862be2a173976ca11
```
bbenligiray, line 19: May as well link this https://github.com/wagmi-dev/viem/blob/main/src/chains/definitions/mainnet.ts

.env.example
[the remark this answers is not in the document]
mcoetzee: @Siegrift FYI the existing variable name on Fleek is `REACT_APP_MAINNET_PROVIDER_URL`, so I guess easiest would be just to rename your `REACT_APP_MAINNET_RPC_URL` to `REACT_APP_MAINNET_PROVIDER_URL`

Siegrift:
> FYI I think there is some more whitelisting to do, because I just ran this locally on mainnet and got some 403s from the provider:

Mhh, strange. I only have the same [whitelisting set as Burak](https://github.com/api3dao/api3-dao-dashboard/pull/431#discussion_r1382879665) and I do not see any errors. I tried clicking a bunch of things and also proposals created by Burak and his ENS name shows correctly. What wallet/app-page did you reproduce that?

mcoetzee:
> Mhh, strange. I only have the same [whitelisting set as Burak](https://github.com/api3dao/api3-dao-dashboard/pull/431#discussion_r1382879665) and I do not see any errors.

It's no longer an issue because Burak has added [0xc0497E381f536Be9ce14B0dD3817cBcAe57d2F62](https://etherscan.io/address/0xc0497E381f536Be9ce14B0dD3817cBcAe57d2F62) to the whitelist 👍

mcoetzee (approved):

Siegrift:
> It's no longer an issue because Burak has added [0xc0497E381f536Be9ce14B0dD3817cBcAe57d2F62](https://etherscan.io/address/0xc0497E381f536Be9ce14B0dD3817cBcAe57d2F62) to the whitelist 👍

But I am using my alchemy instance and I don't have the contract whitelisted. I am just wondering, why I wasn't able to reproduce that error.
image

mcoetzee:
> But I am using my alchemy instance and I don't have the contract whitelisted. I am just wondering, why I wasn't able to reproduce that error. 

I'm guessing those ENS names were already cached for you locally. Try clearing your local storage.

Siegrift:
> I'm guessing those ENS names were already cached for you locally. Try clearing your local storage.

Tried on incognito with both MM and TrustWallet and didn't see those errors.

PR #441, opened 2024-03-23
Bump versions of Actions steps in order to add provenance

dcroote:
"Attestations are supported with version 4 and later of the docker/build-push-action" ([source](https://docs.docker.com/build/ci/github-actions/attestations/)) - this PR updates from v3 to v5, and updates other steps as well.

EDIT: The `production` branch is the only one from which images get pushed because of this bool:

https://github.com/api3dao/api3-dao-dashboard/blob/b8befd30d13bf352c12b2bfd4e0a0b239352c5de/.github/workflows/build-docker.yml#L37

so I originally opened the PR against `production`, but realized that didn't make sense since commits are merged from `main` to `production`. In rebasing, I did notice GitHub is having some trouble reconciling `main` and `production`: see [this compare](https://github.com/api3dao/api3-dao-dashboard/compare/production...main). I'm guessing, based on the [co-authored commits on production](https://github.com/api3dao/api3-dao-dashboard/commits/production/) that a "rebase and merge" was used in #432 instead of using a merge commit.

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -12,12 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
       - name: Docker Metadata
         id: meta
-        uses: docker/metadata-action@v4
+        uses: docker/metadata-action@v5
         with:
           images: api3/dao-dashboard
           tags: |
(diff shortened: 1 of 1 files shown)


Siegrift (approved):
About:

> I did notice GitHub is having some trouble reconciling main and production: see [this compare](https://github.com/api3dao/api3-dao-dashboard/compare/production...main).

In general I prefer rebasing (and linear history). Initially, I wanted to do rebases into the `production` branch. This should just fast-forward the release branch, but GitHub [does not support fast-forward merges](https://stackoverflow.com/questions/60597400/how-to-do-a-fast-forward-merge-on-github), so it will always say "X commits ahead, X behind". I am OK with doing the merge strategy, but it could be a bit more detailed [here](https://github.com/api3dao/api3-dao-dashboard/blob/production/dev-README.md#updating-the-production-deployment).

dcroote:
I agree it's disappointing GitHub doesn't allow for this, but at least the merge strategy, as the lesser of several evils, avoids the branches appearing to diverge.

PR #442, opened 2024-04-08
Provide REACT_APP_MAINNET_PROVIDER_URL env variable when building Docker image

### LINKED ISSUE #439: Cannot run prebuilt docker image
tl;dr: It appears the prebuilt docker image does not have the necessary `REACT_APP_MAINNET_PROVIDER_URL` env variable and it cannot be passed it, but it instead must be included at build time.

## Reproducing

When attempting to run the prebuilt docker image as suggested by the README:

https://github.com/api3dao/api3-dao-dashboard/blob/6aee21b2f668d485f04df75c2cee72e2e5b665ad/README.md#L23-L30

I get a blank screen when connecting to `http://localhost:7770/`. Inspecting the page, there is a console error:
```
wallet-connect.ts:16 Uncaught Error: Missing REACT_APP_MAINNET_PROVIDER_URL env variable
    at 38087 (wallet-connect.ts:16:9)
    at n (bootstrap:19:32)
    at startup:4:82
    at n.O (chunk loaded:25:12)
    at startup:5:43
    at main.fabe0f7b.js:1:146002
```

Providing this env variable either through `--env` or `--env-file` when using the `docker run` command does not solve the issue.

I then tried to build from the `production` branch locally- same result.

I then built from the `production` branch locally with a `.env` file present and that was successful.

mcoetzee:
Closes #439 

### What does this change?
- builds the Docker image with the `REACT_APP_MAINNET_PROVIDER_URL` env variable through the use of a build arg
- updates documentation

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -34,5 +34,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
+          build-args: |
+            REACT_APP_MAINNET_PROVIDER_URL=${{ secrets.REACT_APP_MAINNET_PROVIDER_URL }}
           push: ${{ github.ref == 'refs/heads/production' }}
           tags: ${{ steps.meta.outputs.tags }}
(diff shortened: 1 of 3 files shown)


dcroote (approved):
I tested locally using the updated Dockerfile and instructions and it worked smoothly 👍

PR #444, opened 2024-04-16
Omit encoding target signature and params when creating a proposal with a blank target signature

mcoetzee:


src/logic/proposals/encoding/encoding.ts
```diff
@@ -160,14 +160,21 @@ export const goEncodeEvmScript = async (
...
-      utils.defaultAbiCoder
-        .encode(
-          ['address', 'uint256', 'bytes'],
-          [targetAddress, targetValue, encodeFunctionSignature(targetSignature) + encodedTargetParameters.substring(2)]
-        )
-        .substring(2);
+    const targetCallData =
+      // If this is a version 1 proposal, then we need to keep encoding the target signature regardless if it's blank
+      // or not, otherwise we could get a script mismatch and a false-positive for the malicious proposal check
+      version === '1' || targetSignature
```
Siegrift, line 166: fwiw, I was thinking whether this should be moved to a separate function to better see the versioning complexity, but doing that is a bit random/premature (in the future the next version may alter different part of the code). So I think this fine.

src/logic/proposals/encoding/encoding.test.ts
Siegrift, line 1: I think there should be a test showing that the encoded EVM script differs in v1 and v2 scheme (and we expect the v2 to be shorter).

src/logic/proposals/encoding/encoding.test.ts
```diff
@@ -359,7 +360,8 @@ describe('isEvmScriptValid()', () => {
   });
 
   // The data for this proposal ware created locally.
-  it('returns true also for simple ETH transfers', async () => {
+  it('returns true for simple ETH transfers (version 1)', async () => {
```
Siegrift, line 363: Can we run all of these tests (that have version fixed to `1`) with both version v1 and v2? Right now, all the tests which do not use `newFormData` are tested against version 1.

Siegrift (commented):
> Is this something we should handle better (like updating the message to be broader), or is this scenario unlikely enough to not worry about?

To me that handling is good. I also agree that it's unlikely scenario.

src/logic/proposals/encoding/encoding.test.ts
Siegrift: 👍 Right. I didn't immediately notice that.

I think it would still be valuable as that is the change this PR brings (shorter calldata) and that was the first thing I expected to see when looked at the implementation.

bbenligiray (approved):
For better backward compatibility we could have used v2 only when the function signature is `""` though this is not very important

PR #450, opened 2024-08-07
Extend proposal validation to catch whitespace in target contract signature

mcoetzee:
### What does this change?
- extends the proposal validation logic to return a new validation error message when the target contract signature includes whitespace
- rearranges the validation logic to produce validation messages in the following order (to better align with the field order that the user sees on the screen):
  1. target address
  2. target contract signature
  3. parameters

### DIFF
--- a/cypress/e2e/new-proposal.cy.ts
+++ b/cypress/e2e/new-proposal.cy.ts
@@ -7,42 +7,48 @@ it('new proposal form validation', () => {
   cy.findAllByText('Governance').filter(':visible').click();
   cy.findByText('+ New Proposal').click();
 
-  // Let's create a secondary proposal
+  // Create a secondary proposal
   cy.findByLabelText('Secondary').should('be.checked');
 
   cy.findByText('Create').click();
   cy.findByText('Title must have at least one alphanumeric character').should('exist');
   cy.findByText('Description must have at least one alphanumeric character').should('exist');
-  cy.findByText('Make sure parameters is a valid JSON array').should('exist');
+  cy.findByText('Please specify a valid account address').should('exist');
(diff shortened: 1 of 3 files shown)


cypress/e2e/new-proposal.cy.ts
```diff
@@ -7,42 +7,48 @@ it('new proposal form validation', () => {
...
   cy.findByText('Make sure parameters is a valid JSON array').should('exist');
-  cy.findByLabelText('Parameters').clear().type('[]');
+  cy.findByLabelText('Parameters').type('[]');
   cy.findByLabelText('Target Contract Signature').type('transfer(address, unit256)');
   cy.findByText('Create').click();
 
-  // Let user fix incorrect number of parameters
+  // Remove the whitespace in the target contract signature
+  cy.findByText('Make sure the contract signature contains no whitespace').should('exist');
+  cy.findByLabelText('Target Contract Signature').clear().type('transfer(address,unit256)');
```
Siegrift, line 36: It's a pity that we didn't test just with `transfer(address, uint256)`.

Siegrift (approved):
👍 LGTM

cypress/e2e/new-proposal.cy.ts
mcoetzee: Not sure I follow? FYI the test make this typo deliberately in order to test the:
```ts
  // Fix the typo in the target contract signature
  cy.findByText('Ensure parameters match target contract signature').should('exist');
  cy.findByLabelText('Target Contract Signature').type('{backspace}'.repeat(8) + 'uint256)');
```

Siegrift: I meant that it's a pity we didn't catch this during testing in the past.

PR #458, opened 2024-10-22
Rebrand DAO Dashboard - Colors & fonts setup

### LINKED ISSUE #454: Rebrand DAO Dashboard - Colors & fonts
- Update CSS variables so they contain all the new colors used in the rebranded DAO dashboard. Use the same naming as in the Figma.
- Explore possibilities how to use the new font styles (either with TS like in the Market or with SCSS if possible). 
- Add the new font styles to the DAO Dashboard.
- Create new CSS variables for breakpoints that will be used in the new designs

peterjurco:
Closes #454 

Setting up some basics to avoid conflicts when working on rebranding in parallel.

DAO Dashboard rebranded designs: [Figma](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update)
Design system: [Figma](https://www.figma.com/design/BhnOcskTIGPZ18JkhbCuNQ/%E2%9F%81--Api3-DS---Branding-Update?node-id=266-11414&node-type=frame&m=dev)

### What does this change?
- Adds new color palette as SCSS variables
- Imports new fonts - Karla & Poppins
- Adds font styles used in new designs in form of `@mixin`s. They can be used like this:
```css
@import '../../styles/fonts.module.scss';

.headerTitle {
  @include font-style-16px;
}
```
- Adds new breakpoints that will be used in new designs

In all - colors, font styles, and breakpoints - I used the same names as on the rebranded API3 Market (which tries to use the same names as in the Design system)


Anboias (approved):
LGTM

src/styles/fonts.module.scss
```diff
@@ -0,0 +1,303 @@
+@mixin font-body-3 {
```
mcoetzee, line 1: Hmm I'm not really following this numbering system. Reading the code won't give any real clue as to what CSS will be included (when using these mixins).

I think I would be more confortable with something like:
```scss
.someStyle {
  @include font-body-md
  // or
  @include font-body-2xl
}
```

src/styles/fonts.module.scss
```diff
@@ -0,0 +1,303 @@
...
+  font-family: Karla;
+  font-size: 20px;
+  font-style: normal;
+  font-weight: 400;
```
mcoetzee, line 5: Hmm feels like the font-weight should not be included in these mixins.

src/styles/fonts.module.scss
```diff
@@ -0,0 +1,303 @@
...
+
+@mixin font-codeSnippet-1 {
+  font-family: 'Source Code Pro';
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 150%;
+}
+
+@mixin font-description-3 {
```
mcoetzee, line 113: These description styles seems quite different from another (different font-family etc). What ties them together as descriptions?

src/styles/fonts.module.scss
peterjurco: The numbering system follows the Design system. We adopted the same system in Market for these reasons:

1. For developers it's very convenient. You just click on a text in Figma and there you can see:
image
You immediately know that you should use `@include font-subtitle-5`. 

2. I expect that all frontends will at some point adopt the design system (we have a [repo](https://github.com/api3dao/design-tokens) syncing design tokens from Figma already). It would be easier to adopt it if we were already following the same names as the design system.

src/styles/fonts.module.scss
peterjurco: Why not? 🤔

src/styles/fonts.module.scss
peterjurco: True, something doesn't feel right here. Let me double check

peterjurco: Fixed. Some of the styles were not set correctly

src/styles/fonts.module.scss
mcoetzee: Alright then. Ideally the design system would be a bit more descriptive.

mcoetzee (approved):

PR #200, opened 2021-07-06
Layout with footer

karolko:

PR #466, opened 2024-11-06
Rebrand page layout

### LINKED ISSUE #455: Rebrand DAO Dashboard - Layout
Update the main styles of the DAO Dashboard's layout according to [rebranded designs](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=8676-107801&node-type=section&m=dev). This includes
- navigation
- menu (desktop and mobile)
- logo
- Connect button
- Connected user dropdown
- Footer
- Error reporting bar

peterjurco:
Closes #455 

### What does this change?
It updates the basic page layout to the newest designs including
- Navigation
- Menu (desktop and mobile)
- Logo
- Connect button
- Account dropdown
- Footer
- Error reporting bar

External links in the desktop sidebar menu are a new addition to the DAO dashboard.

### Implementation notes
- Previous implementation uses both `max-width` and `min-width` CSS media queries. I think using only `min-width` is better for long term maintenance and readability so I kept using only `min-width`. 
- I have extended `Button` component (similarly how we have it on the Market) so we are able to use variants and sizes from the [Design system library](https://www.figma.com/design/BhnOcskTIGPZ18JkhbCuNQ/%E2%9F%81--Api3-DS---Branding-Update?node-id=276-21218&node-type=text&m=dev) as they are used in the designs, without the need for styling each button separately. I haven't added all the variants, they can be added as they appear in the designs


peterjurco:
I have styled the Error reporting bar

image

But as discussed on the design call, we will change it to the similar as on the market

image

It will include privacy modal (not yet in the designs) which will display when user clicks on the "Manage Settings".  It's supposed to prepare us for adding Google Analytics and Clarity to the DAO dashboard site. I will create a new task for that.

src/components/menu/menu.tsx
```diff
@@ -53,30 +71,28 @@ export const MobileMenu = () => {
       </div>
       <div className={classNames(styles.mobileMenu, { [styles.open]: open })}>
```
Anboias, line 72: I'm not sure about this screen (around 1000px), it doesn't feel right. We probably need to center it and have a full-width button, but it's just a guess.

I suggest we ask Design Team to provide a screen for the mobile menu on SM breakpoint. I see we don't have this in the previous version of the DAO, thus there's no way for us to tell how it should look like.

image

src/components/menu/menu.module.scss
```diff
@@ -111,37 +111,72 @@
...
+  .externalLink {
+    display: flex;
+    align-items: center;
+    margin-left: 0;
+    margin-bottom: 16px;
+    height: 24px;
+    gap: 16px;
+    text-decoration: none;
+    @include font-heading-9;
+
```
Anboias, line 147: We need a min-width on the svgs

```suggestion
    svg {
      min-width: 24px;
    }
```

image

src/components/menu/menu.module.scss
```diff
@@ -152,35 +187,25 @@
...
-    border-top: 1px solid $secondary-color;
+    border-top: 1px solid $color-dark-blue-300;
   }
 }
 
 .menuMobileItem {
   display: flex;
   align-items: center;
-  margin-bottom: $space-lg;
+  margin-bottom: 8px;
```
Anboias, line 199: Here too

```ts
  svg {
    min-width: 32px;
  }
```

src/components/menu/menu.module.scss
```diff
@@ -1,103 +1,103 @@
...
-  @media (min-width: $min-md) {
-    height: 40px;
-    left: -$space-xxl;
-  }
+.externalLinksHeader {
+  margin: 40px 0 8px 32px;
+  @include font-body-14;
+  color: $color-dark-blue-50;
+  display: flex;
+  align-items: center;
```
Anboias, line 61: I would add these to preserve shape on very small screens

```ts
  white-space: nowrap;
  min-width: 200px;
```

src/components/navigation/navigation.module.scss
```diff
@@ -3,33 +3,47 @@
 .navigation {
   height: $navigation-height;
```
Anboias, line 4: Something is off with the navigation height. It's defined in variables as having 120px for all screens, but that's not true in the new designs. As I see it, the heights should be as such:

| | |
|--------|--------|
| Unconnected Wallet: | XS: 56, SM: 56, MD: 120, LG: 120 |
| Connected Wallet:     | XS: 64, SM: 64, MD: 120, LG: 120 |

In the layout css module, we're computing the min-height using the navigation and footers height. Guys working on the DAO back in the days introduced that [here](https://github.com/api3dao/api3-dao-dashboard/pull/200), but I have no idea why that was needed. I suspect it had to do with the bottom image from the previous design, but I'm not super confident about it.
image

So bottom line is height is wrong, so we should adjust it. I'd say the way to go is remove that calculation and leave the page to compute its own height. If we do this now, at the beginning, we'll discover any issues down the road while implementing the rest of the designs. We can also come back to it if needed.

| Not Connected | Connected |
|--------|--------|
| image | image |

Anboias (changes requested):
Looking good 🤙 
I'm halfway through the review, dropping the comments I have so far to speed things up.

src/components/button/button.tsx
```diff
@@ -1,43 +1,87 @@
...
+export type Size = 'xxs' | 'xs' | 'sm' | 'md' | 'lg' | 'normal' | 'large'; //TODO: remove normal and large after rebranding is done;
+
+type BreakpointsProps = { [key in BreakpointKeys]?: { size?: Size } };
+
+export interface Props extends BreakpointsProps {
   children: ReactNode;
   className?: string;
-  type?: 'primary' | 'secondary' | 'link' | 'text';
-  size?: 'normal' | 'large';
+  type?: 'primary' | 'secondary' | 'link' | 'text' | 'menu-link-secondary' | 'link-blue';
```
mcoetzee, line 13: Hmm I think we are going down a weird path here. The "link" variant for the Button component was to give us an actual button element that has similar styling to a link. I understand a Button component that is more flexible in that it can also render an anchor element that has similar styling to a button.

But now it seems this Button component can be used to give us an anchor element that looks like a link:
```tsx
<Button type="link|link-blue|menu-link-secondary" href="..." ... />
```
^^ This usage does not make sense, as it neither renders a button, or styles the anchor as a button.

I think this `link-blue` and `menu-link-secondary` should be removed here. We've got an `<ExternalLink />` component which should probably be extended.

src/components/menu/menu.module.scss
```diff
@@ -1,85 +1,51 @@
...
 }
 
-.menuActiveItem {
+.navLink {
+  border-left: 2px solid transparent;
+}
+
+.activeNavLink {
   cursor: default;
+  border-left: 2px solid $color-blue-500;
```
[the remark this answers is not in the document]
peterjurco: Yeah, I already wrote [two comments](https://www.figma.com/design/BhnOcskTIGPZ18JkhbCuNQ?node-id=266-11414&m=dev#958849159) regarding this. I believe we have a correct color for `blue-500` but I am not sure they want to use `blue-500` on these places. I will write Leandro directly to clarify this.

src/components/menu/menu.tsx
peterjurco: Yeah, there's no design. I will ask about this

peterjurco: Actually, after some more thinking I adapted the same approach as on the market - the mobile menu has fixed width on those tablet screens. WDYT?

src/components/navigation/navigation.module.scss
peterjurco: > Guys working on the DAO back in the days introduced that https://github.com/api3dao/api3-dao-dashboard/pull/200, but I have no idea why that was needed. I suspect it had to do with the bottom image from the previous design, but I'm not super confident about it.

Good observation. Initially I went with the approach of not disrupting existing layout but I agree with you that calculated height is not needed with the new designs. I have removed it. 


I have adjusted the navigation heights as well. I was putting 120px height for small screen navigation on purpose, because it was correct for the unconnected state ("Connect wallet" button was counting to that height). But you are right that this doesn't hold for the connected state so it needed some adjustment. Hopefully it's ok this time.

src/components/menu/menu.tsx
Anboias: Great call 👌 
Surely that's the only way it makes sense

Anboias: Oh, and since we're doing this we can prevent the page scrolling when menu is open. Like we did on the Market.

```ts
menu.tsx:
<div className={classNames(styles.mobileMenuWrapper, { "mobile-menu-open": open })}>
```
```ts
index.scss:
html:has(.mobile-menu-open) {
  overflow: hidden;
}
```

src/components/menu/menu.module.scss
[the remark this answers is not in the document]
Anboias: Please also drop a comment about the Disconnect dropdown. It seems it has the same background color as the page, so there's no contrast between them and it looks weird. In Figma, dropdown is placed on a white background, but that's not we use.

Figma
image

Implementation
image

Anboias (commented):
Looks good so far, only a few more comments

src/components/button/button.tsx
peterjurco: Yeah, I can see your point here 🤔 My approach is rooted in [the design library](https://www.figma.com/design/BhnOcskTIGPZ18JkhbCuNQ/%E2%9F%81--Api3-DS---Branding-Update?node-id=322-26668&node-type=section&m=dev), where we have both links and buttons categorized simply as buttons. 

I suppose we could have `link-blue` and `menu-link-secondary` as part of the `<ExternalLink />` as you suggested, but my worries are that:
- The `<ExternalLink />` would share lots of similar traits with the `<Button />` - props like `type, theme, size, disabled, href`. Having two components with almost the same `props` would feel repetitive. 
- It would complicate the code in situation like the footer where we have both 
  - links to external sites ("About API3", "GitHub") and 
  - button opening the Error reporting bar ("Error reporting"). 
- All of these ^^ footer "elements" should be rendered the same while being both buttons and links. If you went through designs you would find more situations like this.

If you feel that component that could be both link and the button shouldn't be actually named "Button", maybe we can come up with something more neutral? I feel like this is more about our own organization of the code than the actual HTML output, since in both approaches the result is: an anchor element for links and button element for buttons.

src/components/menu/menu.tsx
peterjurco: Nice touch. Thanks!

src/components/navigation/navigation.module.scss
peterjurco: I put a `min-width` there to prevent the button to collapse like on the screenshot and to keep it true to designs for the 360px screen at the same time. 

As a general comment, given the hard deadline, we might want to focus ourselves more to the bigger things, we don't optimize the app for small screens like this anyway (Although I do appreciate your thorough review 👍)

src/components/menu/menu.module.scss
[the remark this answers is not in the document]
peterjurco: I was actually about to ask about this low contrast issue on the last design call but there wasn't lot of room for it. I will ask about it this week.

src/components/layout/layout.module.scss
```diff
@@ -33,52 +33,45 @@
...
-  display: flex;
-  align-items: center;
-  justify-content: center;
+@media (min-width: $md) {
+  .footer {
+    height: $footer-height-desktop;
 
-  @media (max-width: $max-sm) {
-    flex-direction: column;
+    &Content {
```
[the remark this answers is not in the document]
peterjurco: It's the old class that was there before. Apparently, you can concatenate class names. 

```
<div className={styles.footer}>
<div className={styles.footerContent}>
```

```css
.footer {
   // targets the first div

  &Content {
    // targets the second div
  }
}
```

Anboias (approved):
AFAIC this is good to go after you fix the Disconnect button contrast issue.
Nice job 👍

src/components/button/button.tsx
peterjurco: Ok, I don't want to spend too much time arguing about this, given we have quite a tight deadline now. So let's move to the solution. What do you propose? To remove `link-blue` and `menu-link-secondary` from the `<Button />` and make it part of the `<ExternalLink />` component? How would you implement buttons that have `link-blue` and `menu-link-secondary` style in the designs? There are couple of them - aforementioned "Error Reporting" in the footer and then these `link-blue` ones:

image

image

image

image


If you just don't want to have `<Button />` component rendering as `<a>`  we can use `@mixins` to define `link-blue` style somewhere and reuse it in both `<Button type="link-blue" />` and  `<ExternalLink type="link-blue" />` and we would use ExternalLink in the case it's an external link and button in every other case. Does it work for you?

mcoetzee: > If you just don't want to have <Button /> component rendering as <a> we can use @mixins to define link-blue style somewhere and reuse it in both <Button type="link-blue" /> and <ExternalLink type="link-blue" /> and we would use ExternalLink in the case it's an external link and button in every other case. Does it work for you?

Appreciate the detailed response. Yeah this ^ sounds like a good option, I'm happy to go with it 👍

peterjurco: Done ✅

mcoetzee (approved):
Nice one 🤙

Issue #460, opened 2024-10-31
Rebrand DAO Dashboard - Governance page


peterjurco:
Update a layout for the Governance page based on the [new designs](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=8676-108078&node-type=frame&m=dev). This includes:
- Skeletons
- Treasury dropdown 
- Empty Active Proposals table 
- Wallet disconnected state
- All the tooltips on the page

PrenaamD:
@peterjurco it seems that the delegate tooltip does not conform with design
Implementation

https://github.com/user-attachments/assets/02a8e711-6d0e-443f-961c-9279981dc14d


Design

![Image](https://github.com/user-attachments/assets/d066d7a2-2164-4b1e-a930-690412877914)

peterjurco:
@PrenaamD I'm not sure why designs say that but in the DAO dashboard we are not supposed to change any logic. We should just style the current implementation. So if the tooltip in the old implementation said 
```
You have staked API3 tokens.
You haven't updated delegation in the last 7 days.
```

I am not changing it unless I am explicitly told.

PrenaamD:
Ok @peterjurco makes sense. I will check with the design team.

PR #496, opened 2024-11-20
Rebrand Governance page

peterjurco:
Closing #460 

### What does this change?
- Rebrand the Governance page, excluding proposals table & delegation modals (they will be done in separate tasks)

image


Anboias (approved):
Implementation looks good to me, nice job 👌  

I left some comments about the Button wrapper change I've done recently, so I can come back to this PR later and see where to adjust.

Besides that, I noticed two things that are bugging me. First one is about the page title. I'm not sure how I overlooked this when doing the Staking page, but I strongly feel that it shouldn't be a paragraph. I'll kindly ask you to make it a `h1`.

image

Second one is about the `Delegate` button. It isn't so _obviously_ disabled as in the previous design, and I found myself rage-clicking it for a bit. At least we could have a reason for it being disabled stated in the tooltip. Wdys, is it just me? If not, we can bring it up in the next week's Design call.

https://github.com/user-attachments/assets/65ecf4cf-7b03-492d-b96d-14212eb1c66b

peterjurco:
> I strongly feel that it shouldn't be a paragraph. I'll kindly ask you to make it a h1.

I totally agree. Updated. 


> Second one is about the Delegate button. It isn't so obviously disabled as in the previous design, and I found myself rage-clicking it for a bit.

Yeah, me as well. I was about to drop a comment in the Figma but forgot. Thanks for reminding me.

mcoetzee (approved):

PR #500, opened 2024-11-25
Rebrand New Proposal modal

### LINKED ISSUE #464: Rebrand DAO Dashboard - New proposal
Update "New proposal" modal on the Governance page based on the [new designs](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=8676-108150&node-type=instance&m=dev).

Image

Anboias:
Closes #464

## What does this do?
Updates New Proposal modal to new branding

## How did I action this?
* I split the existing `Input` component in two (`AutosizeInput` and `TextareaInput`), because the input from other modals and input from this one diverged too much. Felt like there was no point in overcrowding all the logic into a single file

* Designs only show us two screens for the Modal: SM and LG. Had to fill the gaps myself, so I did some small adjustments in terms of spacings and sizings for MD (between 768 and 1200) to make things fit better

#### SM
image 

#### MD
image

#### LG
image


src/pages/dashboard/forms/token-deposit-form.tsx
```diff
@@ -95,7 +95,7 @@ const TokenDepositForm = (props: Props) => {
 
       <div className={styles.formContent}>
         <div className={styles.inputWrapper}>
-          <Input type="number" autosize value={inputValue} onChange={(e) => setInputValue(e.target.value)} autoFocus />
+          <AutosizeInput type="number" value={inputValue} onChange={(e) => setInputValue(e.target.value)} autoFocus />
```
peterjurco, line 98: I would keep the old component name - `<Input />`. We are using `react-input-autosize` library under the hood but from the outside it's just input.

peterjurco (approved):
Looks good. Thanks for filling up the missing design and sensitive refactoring that improves the code but it's not overwhelming.

src/components/input/textarea-input.tsx
```diff
@@ -0,0 +1,90 @@
...
+const Input = ({ type = 'text', multiline, ...componentProps }: Props) => {
+  const CustomNumberInput = useCallback((props: any) => {
+    return (
+      <div className={styles.input}>
+        <input {...props} />
+      </div>
+    );
+  }, []);
+
+  if (multiline) {
```
mcoetzee, line 36: Hmm I'm a bit confused. Can we extract the textarea back into its own `<Textarea />` component? I don't see the need to combine it with other things. Top component libs like shadcn, headless UI, react-aria keeps it separate. Also the name `<TextAreaInput />` can be misleading as it almost sounds like it should always render a textarea. I wouldn't naturally reach for it for a standard input.

src/components/input/input.tsx
mcoetzee, line 1: You mentioned in the PR description that you split the `input.tsx` up in the `AutocompleteInput` - did you mean to say `AutosizeInput`? If so can you rename this file accordingly.

Following previous comment, I think it makes sense to have:
- `<Input />`
- `<AutosizeInput />`
- `<Textarea />`

src/components/input/textarea-input.tsx
Anboias: It made sense for me in our case because it's aligned with the designs this way. We have three different inputs in the designs. It feels to me that 1 & 2 (not implemented yet) make sense to share the same component, while 3 can be a separate one.

### 1. Deposit/Stake/Unstake/...etc. modals ([design](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=2373-67360&m=dev))
- Called `FormField` in designs
- Implemented. Uses Autosize

image

### 2. Delegate modal ([design](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=8272-117999&m=dev))
- Called `FormField` in designs (but different group)
- Not yet implemented. Simple, no Autosize

image

### 3. New Proposal modal ([design](https://www.figma.com/design/BhnOcskTIGPZ18JkhbCuNQ/%E2%9F%81--Api3-DS---Branding-Update?node-id=890-6963&m=dev))
- Called `Input` in designs
- Implemented in `TextareaInput.tsx`

image

Anboias: >  I don't see the need to combine it with other things. Top component libs like shadcn, headless UI, react-aria keeps it separate.

I actually see the point. There are two arguments I have for this approach:
A) We are aligned with the designs. Both `input` and `textarea` are similar and part of the same component in Figma, so they could share the styling in code as well

B) We don't need to be as modular as those libraries since our use case is very specific

src/components/input/input.tsx
Anboias: So following your suggestion, to have `<Input />`, `<AutosizeInput />` and `<Textarea />`, I would like to suggest a different approach, considering that `Input 2` needs to be implemented as well:

* We rename `<AutosizeInput />` back to `<Input />`. Add a `autosize` prop and combine `Input 1` and `Input 2`
* We rename `<TextareaInput />` to `<FormInput />` or similar. Allow it be simple or multiline

If you don't agree with the idea of keeping `textarea` together with `input`, I suggest we do this instead:

* We rename `<AutosizeInput />` back to `<Input />`. Add a `autosize` prop and combine `Input 1` and `Input 2`
* We separate `<TextareaInput />` to `<TextareaInput />` and `<SimpleInput />` or similar

Wdyt?

src/components/input/textarea-input.tsx
mcoetzee: Ok, let's name this `<Input />` then.

src/components/input/input.tsx
Anboias: As discussed in private, I went for separating the components after all 😄 
That was a quick adjustment and it feels right to have them like this. Thanks for helping me come to this outcome.

This is the new structure:
```
input/input.tsx
form/input.tsx
form/textarea.tsx
```

mcoetzee (approved):
🤙

PR #505, opened 2024-11-27
Update account dropdown 

peterjurco:
The designs for the account dropdown were slightly adjusted, mainly the background but also some spacings and sizes.

Designs: [Figma](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=8676-108481&node-type=instance&m=dev)

| Before | After |
|--------|-------|
| image | image |

### DIFF
--- a/src/components/sign-in/sign-in.module.scss
+++ b/src/components/sign-in/sign-in.module.scss
@@ -1,9 +1,12 @@
 @import '../../styles/variables.module.scss';
 @import '../../styles/fonts.module.scss';
 
+.accountDropdownOpen {
+  background: linear-gradient(180deg, rgba(250, 250, 250, 0.7) 1.5%, rgba(227, 234, 255, 0) 103.92%), #e3eaff;
+}
+
 .accountDropdown {
   margin-top: 20px;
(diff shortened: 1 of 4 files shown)


src/components/sign-in/sign-in.module.scss
```diff
@@ -3,27 +3,34 @@
 
 .accountDropdown {
   margin-top: 20px;
+  background: linear-gradient(180deg, rgba(250, 250, 250, 0.7) 1.5%, rgba(227, 234, 255, 0) 103.92%), #e3eaff;
```
Anboias, line 6: Looks like the button background should to be applied only when the dropdown is expanded

| Figma | Implementation |
|--------|--------|
| image | image |

Anboias (changes requested):

src/components/sign-in/sign-in.module.scss
peterjurco: oh ofc, thanks 👍

Anboias (approved):

PR #508, opened 2024-11-29
Add cookie consent logic

### LINKED ISSUE #467: Add Cookie consent modal
As preparation for adding Google Analytics and Clarity to the DAO dashboard we want to add user consent for analytical cookies (along the error reporting ones) to the DAO Dashboard.  Clicking on "Manage Settings" user should see the Privacy settings modal similar to the one on the Market. This should apply only to the new rebranded version. 

Design:  https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=9704-161128&node-type=frame&m=dev

Image

Image

Anboias:
Closes #467 

## What does this do?
- adapts existing consent banner
- implements consent logic from Market
- adds Privacy Settings modal
- adds Checkbox component

## Screenshots

| Banner |
|--------|
| image | 

| Modal |
|--------|
| image |

### DIFF
--- a/src/components/external-link/external-link.tsx
+++ b/src/components/external-link/external-link.tsx
@@ -1,4 +1,4 @@
-import { ReactNode, useEffect } from 'react';
+import { KeyboardEvent, MouseEvent, ReactNode, useEffect } from 'react';
 
 interface Props {
   className?: string;
@@ -22,8 +22,19 @@ const ExternalLink = (props: Props) => {
     }
   }, [href, props.href]);
 
+  const handleEvent = (event: KeyboardEvent<HTMLAnchorElement> | MouseEvent<HTMLAnchorElement>) => {
+    event.stopPropagation();
+  };
+
   return (
-    <a href={href} className={className} target="_blank" rel="noopener noreferrer">
+    <a
+      href={href}
+      className={className}
+      target="_blank"
+      rel="noopener noreferrer"
+      onKeyDown={handleEvent}
+      onClick={handleEvent}
+    >
       {children}
     </a>
   );
--- a/src/components/external-link/external-link.tsx
+++ b/src/components/external-link/external-link.tsx
@@ -1,4 +1,4 @@
-import { KeyboardEvent, MouseEvent, ReactNode, useEffect } from 'react';
+import { ReactNode, useEffect } from 'react';
 
 interface Props {
   className?: string;
@@ -22,19 +22,8 @@ const ExternalLink = (props: Props) => {
     }
   }, [href, props.href]);
 
-  const handleEvent = (event: KeyboardEvent<HTMLAnchorElement> | MouseEvent<HTMLAnchorElement>) => {
-    event.stopPropagation();
-  };
-
   return (
-    <a
-      href={href}
-      className={className}
-      target="_blank"
-      rel="noopener noreferrer"
-      onKeyDown={handleEvent}
-      onClick={handleEvent}
-    >
+    <a href={href} className={className} target="_blank" rel="noopener noreferrer">
       {children}
     </a>
   );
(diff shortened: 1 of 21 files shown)


src/components/layout/error-reporting-notice/error-reporting-notice.tsx
```diff
@@ -1,66 +1,73 @@
...
+    localStorage.setItem(ALLOW_ERROR_REPORTING, allowReporting.toString());
+    localStorage.setItem(ALLOW_ANALYTICS, allowAnalytics.toString());
+
+    if (allowAnalytics) {
+      initAnalytics();
+      (window as any).clarity?.('consent');
+    }
+
+    if (allowReporting) {
+      // initSentry(); // TBD later
```
peterjurco, line 30: We should continue to report errors to sentry. Now we won't be doing it because in `src/index.tsx` we are still using the old  local storage key ("reportErrors")

peterjurco (changes requested):
Design looks good 👌 But we need to use the new local storage key for initializing the sentry.

src/components/layout/error-reporting-notice/error-reporting-notice.tsx
Anboias: Oh, yeah. My bad.
It's fixed now. Took the initialisation logic from the Market for both analytics and sentry

src/utils/error-reporting.ts
```diff
@@ -0,0 +1,25 @@
+import * as Sentry from '@sentry/react';
+
+let initialised = false;
+export const initSentry = () => {
+  if (initialised) return;
```
peterjurco, line 5: Previously there was also condition checking if `process.env.REACT_APP_SENTRY_DSN` is defined. Maybe it can be added here too.

peterjurco (approved):
LGTM

src/components/layout/privacy-settings-modal/privacy-settings-modal.tsx
```diff
@@ -0,0 +1,78 @@
...
+            Visit our Privacy Policy
+          </ExternalLink>
+          <img src={images.externalLink} alt="" className={styles.externalLinkIcon} />
+        </div>
+
+        <div className={styles.checkboxes}>
+          <div className={classNames(styles.checkboxContainer, { selected: allowReporting })}>
+            <CheckBox checked={allowReporting} onChange={setAllowReporting} label="Allow error reporting">
+              <div className={styles.sentryButton}>
+                <ExternalLink href={links.SENTRY}>Sentry</ExternalLink>{' '}
```
mcoetzee, line 47: Hmm by the looks of it, clicking this link/pressing enter or space while it is focused will check the checkbox? We should prevent that if it is the case.

mcoetzee (changes requested):

src/components/external-link/external-link.tsx
```diff
@@ -22,8 +22,19 @@ const ExternalLink = (props: Props) => {
     }
   }, [href, props.href]);
 
+  const handleEvent = (event: KeyboardEvent<HTMLAnchorElement> | MouseEvent<HTMLAnchorElement>) => {
```
mcoetzee, line 25: We don't want to do this globally. You've got a rather weird use case (a link inside an interactive element), so it should be handled there rather.

Anboias: I see. Done.
I had to extend ExternalLink to accept the anchor props and also took this opportunity to clean the `href` parsing logic a bit

src/components/external-link/external-link.tsx
```diff
@@ -1,32 +1,37 @@
-import { ReactNode, useEffect } from 'react';
+import { ComponentPropsWithoutRef, useEffect } from 'react';
 
-interface Props {
-  className?: string;
+interface Props extends ComponentPropsWithoutRef<'a'> {
```
mcoetzee, line 3: One last thing, we want to omit the `target` or `rel` props (because we don't want to be able to change that for an ExternalLink):

mcoetzee (approved):
One minor comment left, so I'm approving so long 👍

PR #510, opened 2024-12-04
Add 404 page

### LINKED ISSUE #504: Rebrand DAO Dashboard - Add 404 page
Add new 404 page to the DAO dashboard 

[Designs](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=8676-108435&node-type=section&m=dev)

Anboias:
Closes #504 

## What does this do?
Implements 404 page

| Mobile | Desktop |
|--------|--------|
| image | image |


src/pages/not-found/not-found.module.scss
```diff
@@ -0,0 +1,74 @@
...
+    width: 100%;
+    max-width: 310px;
+  }
+
+  @media (min-width: $sm) {
+    gap: 64px;
+    margin-top: 0;
+
+    img {
+      max-width: 460px;
```
Anboias, line 24: I made this value up so that it looks good. We don't have designs for this screen

src/pages/not-found/not-found.module.scss
```diff
@@ -0,0 +1,74 @@
...
+
+.wrapper {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  gap: 48px;
+  height: 100%;
+  width: 100%;
+  margin-top: -88px;
```
Anboias, line 12: This is an offset to overwrite the gap caused by the empty Layout title

peterjurco: Maybe it's worth to write this comment also to the file itself?

src/pages/not-found/not-found.module.scss
peterjurco: IDK. I would rather do `max-width: unset` so the tablet looks the same as desktop. It doesn't overflow or break anything.

Alternatively, if you think it's too big for that screen, we can keep this max-width value but then make the font-size also a bit smaller so we are consistent with scaling the page.

peterjurco (approved):
LGTM, just a couple of comments

src/pages/not-found/not-found.module.scss
Anboias: To me it looks pretty good, not sure why you don't like it.
Let me ask Design Team about it

peterjurco: It's ok to scale the graphic down on a tablet screen but why then keep the title and subtitle the same size?

peterjurco: But it's a really minor thing. I am ok even with the current thing (that's why I approved the PR)

Anboias: Yeah, it's no big deal. But I already dropped the question so I'll quickly adjust to anything they want. Maybe they'll give us adjusted text sizes too

src/pages/not-found/not-found.tsx
```diff
@@ -0,0 +1,27 @@
+import Layout from '../../components/layout';
+import styles from './not-found.module.scss';
+import { images } from '../../utils';
+import Button from '../../components/button';
+
+const NotFoundPage = () => {
+  return (
+    <Layout title="">
```
mcoetzee, line 8: Hmm it's not great to provide an empty string here as it will result in the page having an empty `<h1>`. Probably makes sense to make the title an optional prop.

src/pages/proposal-commons/proposal-details/proposal-details.tsx
```diff
@@ -79,9 +78,14 @@ interface RouterParameters {
 const ProposalDetailsPage = () => {
   useScrollToTop();
   const { typeAndVoteId } = useParams<RouterParameters>();
+  const { push } = useHistory();
   const decoded = useMemo(() => decodeProposalTypeAndVoteId(typeAndVoteId), [typeAndVoteId]);
 
-  if (!decoded) return <NotFoundPage />;
+  if (!decoded) {
+    push('/404');
```
mcoetzee, line 85: Pretty sure we want to keep the same behaviour. That is, to stay on the same URL but render the Not Found page instead of navigating to another URL. FYI we don't have `/404` route.

mcoetzee (changes requested):

src/pages/proposal-commons/proposal-details/proposal-details.tsx
Anboias: Hmm... I thought we may want to adapt to the new 404 page, which is more than just a `Not Found` text now. It`s a page on its own, that has a layout around it. 

I see your point, thought. But if we want to keep the URL `not-found.tsx` needs to be split and to export two components: content with layout and content without layout. The `proposal-details.tsx` doesn't need a layout, it already has one.

But this approach seems a bit unusual to me. Do we really want to do it? I feels like the way to go is redirecting everything that is not found to `/404` and NotFound page. This would solve your final comment too, with the missing route. Wdyt?

image

Anboias: Or maybe we just want a Not Found text in there, instead of the entire 404 page content?

mcoetzee: So this is the `<ProposalDetailsPage />` component:
```tsx
          <Route path="/governance/:typeAndVoteId" exact>
            <ProposalDetailsPage />
          </Route>
```
It returns the whole page (including the layout), so you can return your `<NotFoundPage />` (which will have its own layout).

Anboias: Oh, I see now. Not sure where I messed up that it put layout inside layout for me during development.
Fixed

src/pages/not-found/not-found.tsx
Anboias: Actually I see now that we may always need to have a title, so that we pass it inside Helmet in `layout.tsx`. So it's probably best I pass a flag to hide the h1 whenever not needed

```ts
      <Helmet>
        <title>{`API3 DAO | ${title}`}</title>
      </Helmet>
```

src/pages/not-found/not-found.module.scss
Anboias: Removed the offset entirely after implementing Markus' feedback with the Layout `title`

src/pages/not-found/not-found.module.scss
Anboias: Design Team didn't answer yet. I'll leave it like it is and we can come back to it later if needed

src/pages/not-found/not-found.tsx
```diff
@@ -0,0 +1,27 @@
...
+        <img src={images.notFound} alt="404" className={styles.image} />
+
+        <div className={styles.content}>
+          <div className={styles.text}>
+            <h1>Uh oh, that page doesn’t exist.</h1>
+            <p>Let’s get you back to home.</p>
+          </div>
+
+          <Button href="/" type="primary" className={styles.button} size="sm" sm={{ size: 'lg' }}>
+            Explore data feeds
```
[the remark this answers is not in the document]
Anboias: Updated to `Staking`
image

src/components/layout/layout.tsx
```diff
@@ -13,11 +13,16 @@ type Props = {
   children: ReactNode;
   title: string;
   subtitle?: string;
+  hideHeader?: boolean;
```
mcoetzee, line 16: One last minor thing: we don't actually need this `hideHeader` flag, you can just use the `<BaseLayout />` for your NotFound page:
```ts
<BaseLayout title="404 Not Found">
  ...
```

mcoetzee (approved):
Nice one for the `title` `subtitle` change. Just got one last comment so I'm approving so long.

src/components/layout/layout.tsx
Anboias: Nice one

Issue #506, opened 2024-11-28
Rebrand DAO Dashboard


peterjurco:
### Debts

- [x] Replace the Help tooltip in the New Proposal modal with an external link to docs ([PR](https://github.com/api3dao/api3-dao-dashboard/pull/515))
- [x] Remove code temporarily added in [#499] and [#502] ([PR](https://github.com/api3dao/api3-dao-dashboard/pull/517))
- [x] Fix console errors ([PR](https://github.com/api3dao/api3-dao-dashboard/pull/516))
- [x] Update text in the modals ([UX/UI task](https://github.com/api3dao/tasks/issues/1535))
- [x] Update design of the "This proposal is potentially malicious." warning ([design](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=10535-221092&node-type=frame&t=m5VXclWhl9K97LvM-0), [PR](https://github.com/api3dao/api3-dao-dashboard/pull/518))
- [x] Update styles of unconnected state of the Proposal details page ([UX/UI task](https://github.com/api3dao/tasks/issues/1533))
- [x] Adjust Vote states after they are updated in the design ([UX/UI task](https://github.com/api3dao/tasks/issues/1546))
- [x] Remove unused code (unused icons, CSS variables, BorderedBox component ... )
- [x] Connect Wallet button might be fitted into the header ([UX/UI task](https://github.com/api3dao/tasks/issues/1554))
- [x] Fix 404 page breakpoints ([design](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=10647-109961&t=3NwQckIt9dmDzYfu-1))

peterjurco:
@PrenaamD The every point from the ticket was done in a separate PR, all of them mentioning this ticket so you should see it here ⤴

PrenaamD:
@peterjurco it seems that the "voted for/against" icon is not displayed

![Image](https://github.com/user-attachments/assets/717e347d-79c1-45b9-8c9c-ff2a5754f150)

I am testing locally by reusing the code from this commit [Add testing capabilities](https://github.com/api3dao/api3-dao-dashboard/pull/502/commits/61437faf266b71972cefc4f98948ae7bf53fc9f4)

I will be posting feedback progressively

PrenaamD:
[...]

PR #515, opened 2024-12-05
Change Help tooltip to external link in the New Proposal modal

peterjurco:
First point from the #506 

### What does this change? 
- see title

### Context
- [Designs](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=8676-108151&node-type=frame&m=dev)
- [Discussion](https://api3workspace.slack.com/archives/C06SH3YFMK7/p1732700734669899)

### Screenshot
image

### DIFF
--- a/src/components/modal/modal.module.scss
+++ b/src/components/modal/modal.module.scss
@@ -132,7 +132,6 @@
 
     &Large {
       justify-content: center;
-      padding-left: 64px;
 
       h5 {
         @include font-heading-6;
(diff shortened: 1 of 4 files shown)


Anboias (changes requested):
Looking good, but the horizontal spacing is off. Now that you made the button absolute, we want a right padding equal to the one from the left so that title is kept centered

| Figma |
|--------|
| image | 

| Implementation |
|--------|
| image |

mcoetzee (approved):

PR #519, opened 2024-12-31
Rebrand toast messages

### LINKED ISSUE #503: Rebrand DAO Dashboard - Toast notifications
Update styles of the toast messages based on the [rebranded designs](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=2218-3596&node-type=frame&m=dev)

Anboias:
Closes #503 

### What does this change?
- Adds new variant to Button component
- Rebrands the toast messages in the DAO dashboard app to match [designs](https://www.figma.com/design/NG6QClnVrvPRZ5gJ3s0zen/%E2%9F%81--Api3-DAO---Branding-Update?node-id=10710-181899&m=dev)
- Update Toast behaviour to match Market (responsiveness + props: `limit={3} pauseOnFocusLoss={false} autoClose={5000} closeOnClick`)
- Updates toast messages based on the feedback from Tamara: 
image

### How to test this?

Use buttons from Staking page. Will remove them before merging this PR. Cypress tests are failing because of them.
Preview URL: https://503-rebrand-toast-messages-new-preview-api3-dao-dashboard-dev.on-fleek.app/#/

image


src/components/notifications/clear-storage-button.tsx
Anboias, line 1: Logic taken from [storage-full-notification.tsx](https://github.com/api3dao/api3-dao-dashboard/pull/519/commits/1bec60dc30339b4e5072a61a04f68b39f4955564#diff-8c57aa9eb3f2560b9801abbb642ba33c8b46e4af063aebf12ebf151f9702eb65). I adapted it to a single button that shares the styling with the `View transaction` button

mcoetzee: Hmm I don't understand why `the storage-full-notification.tsx` was refactored? I would have expected that you could just import the styles that you want to apply from a shared module? 

```ts
    notifications.warning({
      message: 'We have detected that your local storage is full. Would you like to clear it and refresh the page?',
      isClearStorage: true,
    });
```
It doesn't feel right having this very specific button as a global flag that you can show for any notification.

mcoetzee:
The "Clear Storage" button is a primary call to action, so I don't think we want to give it the subdued grey text styling. This warning shows up when the app effectively stops working, and that button fixes the app, so it needs to be prominent. 

Screenshot 2025-01-07 at 12 47 17

src/app.tsx
```diff
@@ -113,17 +112,10 @@ window.localStorage.setItem = (key, value) => {
   try {
     setStorageItem(key, value);
   } catch (e) {
-    notifications.warning(
-      { message: <StorageFullNotification /> },
-      {
-        toastId: 'storage-warning',
-        bodyClassName: 'cursor-auto',
-        closeButton: false,
```
mcoetzee, line 121: These properties still apply. The user should not be able to close this notification and it should not auto hide.

Anboias: I followed the designs, but I'll revert to this behaviour.

src/components/notifications/clear-storage-button.tsx
Anboias: > Hmm I don't understand why the storage-full-notification.tsx was refactored?

It felt right given the new structure. It also removed some unnecessary code, such as the storage-full-notification scss module file.

Please note that changes are minimal, I only took away everything that wasn't related to the action button itself.

> It doesn't feel right having this very specific button as a global flag that you can show for any notification.

You're right. I changed it now so that instead of sending the flag, we pass the ClearStorageButton itself. This means that the only difference between this approach and the previous one is that we send the button and text separately.  Wdyt?

mcoetzee: > I changed it now so that instead of sending the flag, we pass the ClearStorageButton or any other action whenever needed

Yeah that's a better fit thanks 👍

T9991:
> The "Clear Storage" button is a primary call to action, so I don't think we want to give it the subdued grey text styling. This warning shows up when the app effectively stops working, and that button fixes the app, so it needs to be prominent.
> 
> Screenshot 2025-01-07 at 12 47 17

@mcoetzee then it sounds like this would make more sense to be an error message instead of a warning. Is that right?

mcoetzee:
> @mcoetzee then it sounds like this would make more sense to be an error message instead of a warning. Is that right?

Yeah I'm happy to convert it to an error message 👍

Anboias:
@T9991, do we also want to use a different Button variant?

T9991:
@Anboias primary blue button

T9991:
If we add a cancel button, what would it do?

Anboias:
> If we add a cancel button, what would it do?

It just closes the toast.

Anboias:
@mcoetzee, Design Team agreed with the change above, their only request was to make the progress bar red. My [latest commit](https://github.com/api3dao/api3-dao-dashboard/pull/519/commits/83d3aff260481fcb6c78c0d4a72a2e69a73d41ba) handles that by setting the progress bar color to the appropriate type in case  `autoClose` is set to false.
 
image

src/components/notifications/notifications.tsx
```diff
@@ -1,86 +1,121 @@
...
+    </button>
+  ),
 };
 
 // NOTE: toasts are throttled to prevent duplicate notifications being displayed.
 // This can occur due to callbacks being fired multiple times in quick succession
 export const info = throttle(
   (props: ToastProps, overrides?: ToastOptions) => {
-    return toast.info(<CustomToast {...props} type="info" />, { ...BASE_OPTIONS, ...overrides });
+    return toast.info(<CustomToast {...props} {...overrides} type="info" />, { ...BASE_OPTIONS, ...overrides });
```
mcoetzee, line 96: Hmm any reason for change in spreading the `overrides`/options as props?

src/pages/dashboard/dashboard.tsx
```diff
@@ -42,9 +51,93 @@ const Dashboard = () => {
   const canWithdraw = tokenBalances && tokenBalances.withdrawable.gt(0);
   const canInitiateUnstake = data && !pendingUnstake && data.userStaked.gt(0);
 
+  const showErrorToasts = () => {
```
mcoetzee, line 54: Remember to remove all the test code.

src/components/notifications/notifications.tsx
Anboias: Yeah, I've done it so that `CustomToast` can get the `autoClose` prop, which is passed as a `ToastOptions` (second argument), not as a `ToastProps` (first argument). See [here](https://github.com/api3dao/api3-dao-dashboard/blob/3c08e1c0139347dfd2773e33b269fc702a27fd21/src/app.tsx#L126). We need to pass it that way so that it ends up as an override.

And because `CustomToast` already expected `ToastOptions`, I felt like spreading is the cleanest approach. But we can also do something like this `{...props} autoClose={overrides.autoClose}`, either way would work. If you see a cleaner way, please shoot 🙏 

image

src/pages/dashboard/dashboard.tsx
Anboias: Done

PR #525, opened 2025-01-17
Move connect button to the navigation bar on small screens

peterjurco:
Part of #506 

### What does this change?
- Moves the connect button to the top bar on screens with width under 1200px and shorten the label from "Connect Wallet" to just "Connect"

| Before | After |
|---------|------|
|image| image

### DIFF
--- a/src/components/sign-in/sign-in.tsx
+++ b/src/components/sign-in/sign-in.tsx
@@ -102,7 +102,7 @@ const SignIn = ({ dark, position }: Props) => {
           md={{ size: 'md' }}
           className={styles.connectButton}
         >
-          <span className={styles.connectButtonLabel}>Connect {isDesktop ? 'Wallet' : ''}</span>
+          <span className={styles.connectButtonLabel}>{isDesktop ? 'Connect Wallet' : 'Connect'}</span>
         </ConnectButton>
       )}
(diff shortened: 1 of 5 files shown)


src/components/sign-in/sign-in.tsx
```diff
@@ -96,11 +98,11 @@ const SignIn = ({ dark, position }: Props) => {
       {!provider && (
         <ConnectButton
           type={dark ? 'primary' : 'secondary'}
-          size="sm"
+          size="xxs"
           md={{ size: 'md' }}
           className={styles.connectButton}
         >
-          Connect Wallet
+          <span className={styles.connectButtonLabel}>Connect {isDesktop ? 'Wallet' : ''}</span>
```
Anboias, line 105: Not a big deal, but we can do without the hanging space after Connect

```suggestion
          <span className={styles.connectButtonLabel}>{isDesktop ? 'Connect Wallet' : 'Connect'}</span>
```

Anboias (approved):
👌

src/components/sign-in/sign-in.tsx
peterjurco: Yeah, that's better, thanks!

PR #533, opened 2025-02-20
Rebrand DAO Dashboard

peterjurco:
Merge the rebranded DAO Dashboard developed & tested on the `rebrand` branch to the `main`


src/pages/proposals/proposals.tsx
```diff
@@ -1,8 +1,8 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
mcoetzee, line 1: Was this meant to be removed?

src/utils/analytics.ts
```diff
@@ -0,0 +1,24 @@
+// import TagManager from 'react-gtm-module';
```
mcoetzee, line 1: Looks like some leftovers to be removed in this file? (I'm referring to all the commented out code in this file)

mcoetzee (approved):
I forgot how much effort this was 🤙. Just got some minor comments, so I'm approving so long.

mcoetzee:
Can you update the links that is failing the CI. I think they should be:
- https://api3-dao-dashboard.on-fleek.app
- https://api3-dao-dashboard-staging.on-fleek.app

And I think they should start showing after the first successful deploy on `main` and `production`

src/utils/analytics.ts
peterjurco: It was supposed to be commented-out code that would be uncommented once we wanted to add google analytics to the DAO Dashboard. But I think it's ok to leave just one comment there

Issue #582, opened 2026-05-21
Connecting with Rabby wallet fails to load proposals


Siegrift:
Steps to reproduce:
1. Go to https://api3.eth.limo/#/governance
2. Connect Rabby wallet
3. The proposals will fail to load

The same behaviour does not happen with Metamask or when connected with the in-browser wallet, which routes the traffic via different RPC URL.

Siegrift:
The problematic part can be boiled down to the following RPC call (callable from dev console):
```
await window.ethereum.request({
  method: 'eth_getLogs',
  params: [{ address: "0xdb6c812e439ce5c740570578681ea7aadba5170b",
             topics: ["0x4d72fe0577a3a3f7da968d7b892779dde102519c25527b29cf7054f245c791b9"],
             fromBlock: '0x0', toBlock: 'latest' }]
})
```

when this request flows through the Rabby RPC URL the response is `undefined`. If it goes through MetaMask, the response is an array of logs. (The easiest way to test this is to only enable a single wallet extension and compare what the dashboard returns).

Siegrift:
It seems the fix needs to be done on the Rabby side https://github.com/RabbyHub/Rabby/issues/3750

Siegrift:
Not yet, there is no response from the Rabby team on the issue.

Siegrift:
The [fix](https://github.com/api3dao/api3-dao-dashboard/pull/583) is working quite well and the upstream issue has been closed. I think we can accept this patch as the final solution.

PR #583, opened 2026-05-21
Use our RPC URL for all wallet providers except MetaMask

Siegrift:
Relates to https://github.com/api3dao/api3-dao-dashboard/issues/582 

This is one way to get rid of the problem for the users. How it was previously implemented made Rabby connection flow the requests via Rabby default RPC which fails to fetch the logs. This PR changes the behaviour such that we only use the signer RPC if the connected wallet is MetaMask.

### DIFF
--- a/src/contracts/hooks.ts
+++ b/src/contracts/hooks.ts
@@ -10,6 +10,19 @@ import {
   TimelockManager__factory as TimelockManagerFactory,
 } from './artifacts/factories';
 import { initialChainData } from '../chain-data/state';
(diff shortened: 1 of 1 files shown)


mcoetzee (approved):
Nice one 🥂

Looks like this `isMetaMask` check luckily gets a bit simpler with wagmi v2 👍 (I've added a comment https://github.com/api3dao/api3-dao-dashboard/issues/558#issuecomment-4506312839)

PR #592, opened 2026-08-17
Add CI to deploy to IPFS/upload to Pinata

mcoetzee:
Closes #557 

The CI emits a summary (Cloudflare vibe) to make it easier to get the relevant information:
Screenshot 2026-08-17 at 17 09 38
E.g. https://github.com/api3dao/api3-dao-dashboard/actions/runs/32041149537

Each run also creates a GH deployment: https://github.com/api3dao/api3-dao-dashboard/deployments

PR #594, opened 2026-08-18
Document Pinata key creation

Siegrift:
I've also set it inside GH actions, which resolves https://github.com/api3dao/api3-dao-dashboard/pull/592#issuecomment-5326861800


mcoetzee (approved):
🥂

